PR #154, opened 2017-06-25
:2924 add wxSTAY_ON_TOP for the message.

enpinion:
If the application is maximized, finish message hides behind the frame occasionally. wxSTAY_ON_TOP to say on top to avoid this situation.

### DIFF
--- a/src/program/BodySlideApp.cpp
+++ b/src/program/BodySlideApp.cpp
@@ -2921,7 +2921,7 @@ void BodySlideFrame::OnBatchBuild(wxCommandEvent& WXUNUSED(event)) {
 
 	if (ret == 0) {
 		wxLogMessage("All sets processed successfully!");
-		wxMessageBox(_("All sets processed successfully!"), _("Complete"), wxICON_INFORMATION);
+		wxMessageBox(_("All sets processed successfully!"), _("Complete"), wxICON_INFORMATION | wxSTAY_ON_TOP );
 	}
 	else if (ret == 3) {
 		wxArrayString errlist;


ousnius:
I couldn't reproduce this issue, the wxSTAY_ON_TOP flag doesn't seem to change anything for me for the better or worse.

Is there a way for you to reliably reproduce it? Which windows do you have open and maximized when you run a batch build?

ousnius:
Enabling the flag causes the preview window to minimize or go behind other windows when the finish message pops up.

enpinion:
I'm using Windows 10 (1703 15063.413 latest).
I'd attach a video too.

1. Start a batch build.
2. bring another window and make it focused.
3. Once the batch process has finished, the msgBox will hide behind of the main window.

[test.zip](https://github.com/ousnius/BodySlide-and-Outfit-Studio/files/1101032/test.zip)

ousnius:
It's not happening for me (the message box is never behind the BodySlide main frame for me). I'm on the latest Windows 10 myself as well.

It must be because of a shell extension you may be using. My Windows 10 is fully unchanged.

The video file is corrupted (it doesn't play correctly, not even in Adobe Premiere).
EDIT: It plays in VLC.

enpinion:
I upload the video on Youtube
https://youtu.be/VKLpOkeI9aQ

ousnius:
I'm sorry, but I can't merge this. The issue is your Windows installation or some sort of extension.

Adding the wxSTAY_ON_TOP flag on my system causes for example the preview window to minimize, so for me it would actually make things worse.

Thank you for the effort, though. :)

PR #194, opened 2018-08-13
Minor housekeeping

andrew-boyarshin:
* Fix never showing error dialog (result not assigned to bool variable)
* Add `nullptr` checks to `malloc` & `calloc` calls
* Fix `std::string` -> `char*` -> `std::string` conversions
* Add `std::move` to reduce data copying
* Catch C++ exception by reference

### DIFF
--- a/src/program/OutfitProject.cpp
+++ b/src/program/OutfitProject.cpp
@@ -1685,7 +1685,7 @@ int OutfitProject::LoadReferenceNif(const std::string& fileName, const std::stri
 		if (IsBaseShape(s)) {
 			std::string newName = s + "_ref";
 			refNif.RenameShape(s, newName);
-			baseShape = newName;
+			baseShape = std::move(newName);
 			break;
 		}
 	}
@@ -1771,7 +1771,7 @@ int OutfitProject::LoadReference(const std::string& fileName, const std::string&
 		if (s == shape) {
 			std::string newName = s + "_ref";
 			refNif.RenameShape(s, newName);
-			shape = newName;
+			shape = std::move(newName);
 			break;
 		}
 	}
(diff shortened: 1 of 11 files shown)


src/program/OutfitProject.cpp
```diff
@@ -1135,19 +1135,19 @@ void OutfitProject::SetTextures(NiShape* shape, const std::vector<std::string>&
 
 			if (!mat.Failed()) {
 				if (mat.signature == MaterialFile::BGSM) {
-					texFiles[0] = mat.diffuseTexture.c_str();
```
ousnius: The c_str() is needed to skip duplicate NULL terminators that were added to the strings from IO. Without the c_str(), the strings will end up at a size of 1 with \0 as the explicit content, which will make my !texFiles[i].empty() check a few lines below fail.

Until I find a way to fix the \0 terminators in IO after reading the material files, this change has to be removed (same in PreviewWindow.cpp). I might need to resize the string to c_str() after reading it from the file.

src/program/PreviewWindow.cpp
```diff
@@ -246,19 +246,19 @@ void PreviewWindow::AddNifShapeTextures(NifFile* fromNif, const std::string& sha
 
 		if (!mat.Failed()) {
 			if (mat.signature == MaterialFile::BGSM) {
-				texFiles[0] = mat.diffuseTexture.c_str();
```
ousnius: Same here.

ousnius (changes requested):
See review comments. Otherwise seems fine, thanks.

andrew-boyarshin:
@ousnius reverted those changes to your request.

ousnius (approved):

PR #197, opened 2018-10-19
Split GameDataPath onto input and output

zeule:
Setting independent output directory makes sense
when using Mod Organizer 2 and allows to keep
generated meshes in their own mods.

### DIFF
--- a/Config.xml
+++ b/Config.xml
@@ -3,7 +3,7 @@
     <TargetGame>-1</TargetGame>
     <WarnMissingGamePath>true</WarnMissingGamePath>
     <BSATextureScan>true</BSATextureScan>
-	<!-- Archives black list -->
+    <!-- Archives black list -->
     <GameDataFiles>
         <Fallout3>Anchorage - Sounds.bsa; BrokenSteel - Sounds.bsa; Fallout - MenuVoices.bsa; Fallout - Meshes.bsa; Fallout - Misc.bsa; Fallout - Sounds.bsa; Fallout - Voices.bsa; PointLookout - Sounds.bsa; ThePitt - Sounds.bsa; Zeta - Sounds.bsa</Fallout3>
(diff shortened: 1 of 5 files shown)


ousnius (changes requested):
Thanks, the changes themselves are fine.

1. All changes to files that aren't the following have to be removed. You can make a new PR if that makes it easier.
res/xrc/Settings.xrc
src/program/BodySlideApp.cpp
src/program/BodySlideApp.h
Config.xml
2. It doesn't matter but I'd like "dpOutputaPath" renamed to "dpOutputPath".
3. Rename "wxFileName gameDataDir" to "dpOutputPath" in OnSettings().
4. The OnSettings() changes have to be merged into OutfitStudio.cpp as well.

ezsh:
Thanks, Visual Studio comiited those files when I wanted her to change a commit message. Sorry, I'll fix the typo and remove the superfluous files. 

As for the Outfit studio, I'm afraid  I don't understand how the change should fit there. Do I need to take a similar way and make it save files under another path? I'm unfamiliar with the Outfit Studio, sorry

zeule:
Thanks, Visual Studio comiited those files when I wanted her to change a commit message. Sorry, I'll fix the typo and remove the superfluous files.

As for the Outfit studio, I'm afraid I don't understand how the change should fit there. Do I need to take a similar way and make it save files under another path? I'm unfamiliar with the Outfit Studio, sorry

ousnius:
> Thanks, Visual Studio comiited those files when I wanted her to change a commit message. Sorry, I'll fix the typo and remove the superfluous files.
> 
> As for the Outfit studio, I'm afraid I don't understand how the change should fit there. Do I need to take a similar way and make it save files under another path? I'm unfamiliar with the Outfit Studio, sorry

OutfitStudio.cpp has a copy of the OnSettings() function because they're separate tools now. You only need to add the read/write for the new config variable there, nothing else.

zeule:
Do I get it right: to preserve the setting value, and only for that, I need to read the string and write back unchanged?

ousnius:
> Do I get it right: to preserve the setting value, and only for that, I need to read the string and write back unchanged?

No, you add the exact same code you added to BodySlideFrame::OnSettings to OutfitStudioFrame::OnSettings. It's the same dialog and the path can be changed there too, even if it's not used (yet).

zeule:
Ah, so! Thanks for the explanation!

zeule:
Done. But due to my very little experience with Windows, I was not able to link the Outfit Studio binary and test it.

ousnius (approved):

ousnius:
Thanks.

zeule:
Thank you!

PR #215, opened 2019-08-04
[NifOpt] Updated NifFile::Optimize to only convert niTriShape to bsTriShape if strictly required

Guekka:
With the current version, NifOpt converts all shapes to BSTriShape, even if it is not necessary. 
With this change, it will only be performed to avoid crashes or if the user enables "bsTriShape".

### DIFF
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -977,8 +977,10 @@ OptResult NifFile::OptimizeFor(OptOptions& options) {
 
 				BSTriShape* bsOptShape = nullptr;
 				//Only converting niTriShape to bsTriShape if required
-                if(options.bsTriShape || shape->HasType<NiTriStrips>() || shape->HasType<bhkMultiSphereShape>()
-		  || options.headParts)
+                		const bool needsOpt = options.bsTriShape || options.headParts || shape->HasType<bhkMultiSphereShape>()
+                                      || shape->HasType<NiTriStrips>() || shape->HasType<NiTriStripsData>()
+                                      || shape->HasType<NiSkinPartition>();
+               			if (needsOpt)
 				{
 					auto bsSegmentShape = dynamic_cast<BSSegmentedTriShape*>(shape);
 					if (bsSegmentShape) {
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -977,7 +977,8 @@ OptResult NifFile::OptimizeFor(OptOptions& options) {
 
 				BSTriShape* bsOptShape = nullptr;
 				//Only converting niTriShape to bsTriShape if required
-                if(options.bsTriShape || shape->HasType<NiTriStrips>() || shape->HasType<bhkMultiSphereShape>())
+                if(options.bsTriShape || shape->HasType<NiTriStrips>() || shape->HasType<bhkMultiSphereShape>()
+		  || options.headParts)
 				{
 					auto bsSegmentShape = dynamic_cast<BSSegmentedTriShape*>(shape);
 					if (bsSegmentShape) {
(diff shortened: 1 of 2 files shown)


ousnius (commented):
There's more that needs to be done to make sure NIFs don't crash in SSE, so I can't merge as is.

1. No NiTriStrips/NiTriStripsData
2. No strips used in NiSkinPartition (regardless of it being NiTriShapeData or not)
3. Some headparts crash when they're not BSDynamicTriShape

lib/NIF/NifFile.cpp
```diff
@@ -976,159 +976,163 @@ OptResult NifFile::OptimizeFor(OptOptions& options) {
...
-				bsOptShape->SetCollisionRef(shape->GetCollisionRef());
-				bsOptShape->GetProperties() = shape->GetProperties();
-				bsOptShape->GetExtraData() = shape->GetExtraData();
-
-				bsOptShape->transform = shape->transform;
-
-				bsOptShape->Create(vertices, &triangles, uvs, normals);
-				bsOptShape->flags = shape->flags;
+				//Only converting niTriShape to bsTriShape if required
+                if(options.bsTriShape || shape->HasType<NiTriStrips>() || shape->HasType<bhkMultiSphereShape>())
```
ousnius: What was the purpose for bhkMultiSphereShape being checked here?

lib/NIF/NifFile.h
```diff
@@ -12,6 +12,7 @@ struct OptOptions {
 	bool headParts = false;
 	bool removeParallax = true;
 	bool calcBounds = true;
+    bool bsTriShape = false;
```
ousnius: Default should be true to have it work the same way as before by default if the new property isn't changed.

lib/NIF/NifFile.cpp
Guekka: According to Zilav, bhkMultiSphereShape is not compatible with SSE.

lib/NIF/NifFile.h
Guekka: Agreed.

Guekka:
Everything should be fixed.
I appreciate that you took the time to look at it, thanks.

lib/NIF/NifFile.cpp
```diff
@@ -977,8 +977,10 @@ OptResult NifFile::OptimizeFor(OptOptions& options) {
 
 				BSTriShape* bsOptShape = nullptr;
 				//Only converting niTriShape to bsTriShape if required
-                if(options.bsTriShape || shape->HasType<NiTriStrips>() || shape->HasType<bhkMultiSphereShape>()
-		  || options.headParts)
+                		const bool needsOpt = options.bsTriShape || options.headParts || shape->HasType<bhkMultiSphereShape>()
```
ousnius: Shapes (NiShape) can never be bhkMultiSphereShape, NiTriStripsData or NiSkinPartition, so the additional checks won't help. It's also not all NiSkinPartition blocks, but only those with strips in them (they can have both triangles and strips or either).

The entire check should be this:
```cpp
const bool needsOpt = options.bsTriShape || shape->HasType<NiTriStrips>() || hasSkinPartitionWithStrips
```

And it should check for numStrips > 0 in any of the partitions inside the NiSkinPartition block.

```cpp
bool hasSkinPartitionWithStrips = false;

auto skinInst = hdr.GetBlock<NiSkinInstance>(shape->GetSkinInstanceRef());
if (skinInst) {
	skinPart = hdr.GetBlock<NiSkinPartition>(skinInst->GetSkinPartitionRef());
	if (skinPart) {
		// loop through skinPart->partitions and check for numStrips > 0
		//hasSkinPartitionWithStrips = true;
	}
}
```

Guekka:
Hum, I see. I have very few knowledge about meshes, I probably should have looked for more information before.
Anyway, give me a few minutes to fix that and this time, we should be good.

Guekka:
I hope this commit will be enough.

PR #228, opened 2020-01-15
Made changes to compile and run on Linux with gcc.

sts1skj:
1.  Created CMakeLists.txt to build with CMake on Linux.  It might also
work on other systems.

2.  stricmp and strnicmp were being used; they are missing from
gcc's standard library implementation.  (This is a well-known and
long-standing bug in gcc.)  So I added src/utils/StringStuff.{cpp,h}
with implementations of case-insensitive string compare functions
and changed all instances of stricmp and strnicmp to use them.

3.  Added StringStuff to the VC project files so hopefully it'll still
build on that tool (which I do not have).

4.  Fixed a variety of minor bugs where the code was not standard
C++, but presumably VC was permissive enough that it compiled anyway.
Hopefully VC will accept these changes; I don't have VC.  The most
common bug was assigning a wxString to a std::string in an old-style
initializer; switching to the new initializer style fixed it.

5.  Added a few missing '#include' statements.

6.  Made some no-effect changes to silence compiler warnings, such as
adding "default: break;" to switch statements where some of the values
were not handled.

7.  gcc is missing acosf, cosf, sinf, atan2f, and fabsf.  (This is
a well-known and long-standing bug in gcc.)  Luckily, these are
just alternative names for existing functions, so I switched to the
other names.

8.  There were many locations where BS&OS tried to open files
with filenames where the path separator character was backslash.
I changed them to use forward slash instead.  This shouldn't affect
Windows execution, because Windows accepts either type of slash as a
path separator.  I tried my best to not let the slash style propagate
into data files, but I may not have been successful.

9.  There were a few locations where filenames were being converted to
lowercase before the file was opened.  As Linux is case sensitive, this
didn't work.  So I changed all of these instances so that the filename
is not converted.

10.  On Linux, the wxWidgets headers "#include <X11/X.h>", which is an
ancient legacy header file containing a large number of poorly-named
"#define" statements.  I feel that this is a bug in wxWidgets.  There
was a name clash with a function name in the FBX SDK.  So I moved the
"#include <fbxsdk.h>" line from FBXWrangler.h into FBXWrangler.cpp
so that FBX and wxWidgets headers would never be included by the
same file.  To keep FBXWrangler working, I needed to move more of
the implementation details from the header into the source file.
This necessitated employing the private-implementation idiom.

11.  The C integer-absolute-value function "abs" was being used in
TriFile.cpp on floating point values.  gcc produced a warning, and
I judged this highly likely to be a bug, so I fixed it by replacing
"abs" with "std::abs".

12.  Linux doesn't have a registry, so I used "#ifdef _WINDOWS"
around the handful of blocks of registry code.

13.  In BodySlideApp.h, there was some peculiar code for doing something
with number-key presses and Windows events.  I used "#ifdef _WINDOWS"
to disable this code for linux.

14.  A few files were missing trailing newline characters.
Linux doesn't like that, so I added them.

15.  Switched from old-style cast to "reinterpret_cast" a few places
in NormalsGenDialog to silence compiler complaints.

16.  "_snprintf_s" was used in a few places.  It's a Windows-only
function.  I replaced it with snprintf, which does the same thing
only better.

17.  The extension-loading code in GLExtensions.{h,cpp} used
wglGetProcAddress, which is Windows-specific.  I couldn't figure out
how to convert the code to Linux, so, for Linux, GL extensions are
now loaded by GLEW.  This new code would probably work fine on other
systems too, but it would introduce a dependence on the GLEW library.

18.  Three OpenGL API calls had capitalization errors.  I fixed them.

After all this, BodySlide and Outfit Studio build and run, apparently
correctly, on my Linux system.

### DIFF
--- a/BodySlide.vcxproj
+++ b/BodySlide.vcxproj
@@ -769,6 +769,7 @@
     <ClInclude Include="src\utils\ConfigurationManager.h" />
     <ClInclude Include="src\utils\Log.h" />
     <ClInclude Include="src\utils\PlatformUtil.h" />
+    <ClInclude Include="src\utils\StringStuff.h" />
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="lib\FSEngine\FSBSA.cpp" />
@@ -830,6 +831,7 @@
     <ClCompile Include="src\utils\ConfigurationManager.cpp" />
     <ClCompile Include="src\utils\Log.cpp" />
     <ClCompile Include="src\utils\PlatformUtil.cpp" />
+    <ClCompile Include="src\utils\StringStuff.cpp" />
   </ItemGroup>
   <ItemGroup>
     <Xml Include="Config.xml" />
--- a/BodySlide.vcxproj.filters
+++ b/BodySlide.vcxproj.filters
@@ -865,6 +865,9 @@
     <ClInclude Include="src\utils\PlatformUtil.h">
       <Filter>Utilities</Filter>
     </ClInclude>
+    <ClInclude Include="src\utils\StringStuff.h">
+      <Filter>Utilities</Filter>
+    </ClInclude>
     <ClInclude Include="lib\NIF\Factory.h">
       <Filter>Libraries\NIF</Filter>
     </ClInclude>
@@ -1930,6 +1933,9 @@
       <Filter>Libraries\NIF</Filter>
     </ClCompile>
     <ClCompile Include="src\utils\PlatformUtil.cpp">
+      <Filter>Utilities</Filter>
+    </ClCompile>
+    <ClCompile Include="src\utils\StringStuff.cpp">
       <Filter>Utilities</Filter>
     </ClCompile>
     <ClCompile Include="lib\NIF\Factory.cpp">
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,113 @@
+cmake_minimum_required(VERSION 3.10)
+
+project(BSOS)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
+find_package(wxWidgets REQUIRED gl core base net xrc adv qa html propgrid)
+find_package(OpenGL REQUIRED)
+find_package(GLEW REQUIRED)
+find_library(fbxsdk fbxsdk PATHS ../fbxsdk/lib/gcc/x64/release)
+
+set(commonsources
+	lib/FSEngine/FSBSA.cpp
+	lib/FSEngine/FSEngine.cpp
+	lib/FSEngine/FSManager.cpp
+	lib/gli/glm/detail/glm.cpp
+	lib/LZ4F/lz4.c
+	lib/LZ4F/lz4frame.c
+	lib/LZ4F/lz4hc.c
+	lib/LZ4F/xxhash.c
+	lib/NIF/Animation.cpp
+	lib/NIF/BasicTypes.cpp
+	lib/NIF/bhk.cpp
+	lib/NIF/ExtraData.cpp
+	lib/NIF/Factory.cpp
+	lib/NIF/Geometry.cpp
+	lib/NIF/NifFile.cpp
+	lib/NIF/Nodes.cpp
+	lib/NIF/Objects.cpp
+	lib/NIF/Particles.cpp
+	lib/NIF/Shaders.cpp
+	lib/NIF/Skin.cpp
+	lib/NIF/utils/Object3d.cpp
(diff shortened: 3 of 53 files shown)


ousnius:
@sts1skj: Thanks for the detailed description of your changes. I've cloned your branch and the MSVC builds are successful. Basic tests are working but I need to test more.

I don't know if the void pointers thing in NormalsGenDialog is even working as intended (esp. on 64-bit builds), both before and after the changes, but it doesn't matter right now as it's not used.

There are one or two warnings I would like to fix. Additionally, I would like two of the forward slash changes reverted (in the .osp project files).

1. Output path is filled with forward slashes now. I don't want it to suddenly change with the next version, so the .osp should still store it as backslashes:
`<OutputPath>meshes/armor/New Outfit</OutputPath>`
2. When adding new slider data, the data link to the .osd is in forward slashes as well (after TEST.osd):
`<Data name="BaseShapeHRtest" target="BaseShapeHR" local="true">TEST.osd/BaseShapeHRtest</Data>`

Another thing in BodySlideApp.cpp is where it adds the .tri path to the NIF files (NiStringExtraData blocks). These strings need to be stored using backslashes in the NIF file (triPath and triPathTrimmed with regex).

Can you please give me permission to push to the PR branch so I can fix the warnings and anything else I might find?

sts1skj:
When you say "give me permission to push to the PR branch", are you referring to the "Allow edits from maintainers" checkbox?  That is already checked.  Please let me know if there's something else I need to do; I've never used github before.

ousnius:
@sts1skj Sorry, I had cloned the wrong branch. It's working.
I've edited my post above with some issues. Do you want to look at them?

sts1skj:
I will try to fix the two forward-slash problems you mentioned.

ousnius:
For BodySlide you will need to look at the `SliderSet::GetOutputFilePath` function and where it's used for the BodySlide build (wxRemoveFile uses it, and it's used for the tri file paths as well).

POMXARK:
Hi ousnius and Jeir. why not add a collision check? to check clipping, the mass of libraries, or is it difficult to implement (bullet3,FCL)??? to handle convergence and automatically adjust armor to the body? Is it worth trying to do this ???
In my large squat, fatal matches in the lower chest are possible, when it will be possible to automate the conversion process, whoever tried, I want to use macros for the program, as done in notepad ++

sts1skj:
I've checked in a set of changes that should fix the slash problem.  Now, pathnames are converted to backslashes when being written to nif, osp, material, and ref templates xml files.  Pathnames read from these files are converted to use the OS's path separator character.  So now internal pathname strings will generally use backslash on Windows and forward slash on all other OSes, except for string-literal pathnames that don't matter, which generally use forward slashes.

I tried to merge in the changes you made, Ousnius, to silence compiler warnings.  Hopefully I did that correctly; git is still somewhat mystifying to me.

I also added a Readme-linux.txt file with build instructions.

I didn't add code for making pathnames in Config.xml always use backslashes.  If you want me to make the changes for that, let me know.

I also examined the code for reading and writing the following file types to see if they store any pathnames, and I didn't find any:
shader file
tri file
tri head file
obj file
fbx file
slider set group xml
slider preset xml
slider categories xml
Please let me know if any of these types of files do contain pathnames or if I missed a file type.

sts1skj:
I need to do more testing.  Something isn't working right.

sts1skj:
This should be ready now.  I've done a lot of testing, and everything I've tested works fine on Linux now.

sts1skj:
Your commit 26c9e13 didn't break anything on the Linux side, as far as I can tell.

ousnius:
@sts1skj: If building a project in BodySlide, textures in the preview, loading -> saving -> loading a project in Outfit Studio works for you, all while keeping the slider data working, then I can merge the PR.

sts1skj:
@ousnius: I added one more tiny change.  Hopefully it has no effect on Windows; it just sets the background color of sliders in Outfit Studio to what it would be anyway.

I've tested your latest commit ec64a13, and everything works on Linux according to the tests you specified.

sts1skj:
@ousnius: I added one more tiny change.  Hopefully it has no effect on Windows; it just sets the background color of sliders in Outfit Studio to what it would be anyway.

I've tested your latest commit ec64a13, and everything works on Linux according to the tests you specified.

ousnius:
I've added another place where the slider pane background color was set. This should be ready to merge now.

sts1skj:
Everything looks good to me.  Go ahead.

ousnius:
@sts1skj Thank you for the work on this!

sts1skj:
Thank you.

PR #231, opened 2020-01-25
Bug fix: slider scale not applied to vertex moves

sts1skj:
When moving a vertex in Outfit Studio with slider-edit on and the slider not set to 1, the changes were not scaled, resulting in distortion.


ousnius:
@sts1skj Does this work for slider values under 0.0 and above 1.0 as well, say -1.0 or 2.0?
(copy/paste -100% or 200% from Notepad to the OS field)

sts1skj:
Yes, it works for slider values under 0 and above 1.

ousnius:
One isn't supposed to edit sliders at values other than 100%, but I guess this at least makes it so doing so anyway doesn't cause any issues.

sts1skj:
I've found it useful to edit sliders at values higher than 100%, because distortions that need to be cleaned up are emphasized.

PR #233, opened 2020-01-26
Bug fix: slider scale doubled on unset reference

sts1skj:
When a shape was unset as a reference, all of its slider diffs were
being doubled.


ousnius:
This only happened after using "Load Reference", it didn't happen when just setting/unsetting references for the one or multiple shapes. Probably because load reference adds the slider data back that "SetBaseShape" is supposed to delete.

This isn't a true fix for this behavior, but the end result is correct so I will merge.

PR #235, opened 2020-01-27
Cleantweak

sts1skj:
I did a lot of data-structure cleaning in TweakBrush.h and TweakBrush.cpp.  I tried my best to not change how anything works but to just rearrange the data, with one exception (item 3 below).  The changes:

1.  Eliminated all use of "delete" and "free" in favor of std::vector and std::unique_ptr.

2.  Moved a few member variables defined in class scope into the functions that use them because their contents were never used again.  (Example: TB_Smooth::b).

3.  Eliminated the code for asynchronously updating normals in another thread.  It was not thread safe.  Doing the normal update synchronously didn't seem to affect performance for me, even with a huge brush on a high-polygon model.  I did notice a flickering of the model before that may have been the result of the unsafe asynchronous code; eliminating this has apparently cured the flickering.  (This is the only intentional change to functionality in this branch.)

4.  Moved some data storage from TweakStroke into a new data structure, ChangeToOutfit.  The new data structure contains just the project-data changes without any of the brush-related stuff.  Ideally, ChangeToOutfit would be the item stored in the undo history instead of TweakStroke, but I was afraid to make that change without a deeper understanding of Outfit Studio's project data model.

5.  Added "const" in a few places.

6.  Eliminated some unused code.

### DIFF
--- a/BodySlide.vcxproj
+++ b/BodySlide.vcxproj
@@ -769,6 +769,7 @@
     <ClInclude Include="src\utils\ConfigurationManager.h" />
     <ClInclude Include="src\utils\Log.h" />
     <ClInclude Include="src\utils\PlatformUtil.h" />
+    <ClInclude Include="src\utils\StringStuff.h" />
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="lib\FSEngine\FSBSA.cpp" />
@@ -830,6 +831,7 @@
     <ClCompile Include="src\utils\ConfigurationManager.cpp" />
     <ClCompile Include="src\utils\Log.cpp" />
     <ClCompile Include="src\utils\PlatformUtil.cpp" />
+    <ClCompile Include="src\utils\StringStuff.cpp" />
   </ItemGroup>
   <ItemGroup>
     <Xml Include="Config.xml" />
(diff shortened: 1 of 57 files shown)


sts1skj:
Upon further testing, it appears that synchronizing the asynchronous normal updates does impact the performance.  I'll add it back in.

ousnius:
@sts1skj This is breaking the brush mirroring (main stroke does nothing but the mirrored stroke does). Compiles on Windows.

Asynchronous normals are definitely needed. I know they're not fully state safe but it works for now.

sts1skj:
I've committed a fix to that brush-mirroring bug.  It was just a typo in commit aad4f39.

ousnius:
@sts1skj Thanks, should be ready to merge but I don't really like the names "ChangeToOutfit" and "ChangeToShape". Could you rename to something along the lines of "TweakStateHolder" and "TweakState" please?

First struct holds multiple states mapped to each mesh. Second struct is a state of one mesh. The words outfit and shape don't really have any meaning in this part of the code.

sts1skj:
I've just changed ChangeToOutfit and ChangeToShape to TweakStateHolder and TweakState, respectively, as requested.

PR #237, opened 2020-01-28
Fixed three bugs in hclapFilter

sts1skj:
1.  hclapFilter was not handling welded vertices properly.  This resulted
in smoothing brushes unsmoothing such vertices instead of smoothing them.

2.  hclapFilter did not correctly implement the HC Laplacian filter.
It was using the previous iteration's values where it was supposed to use
the original values (where multiplied by hcAlpha).

3.  hclapFilter was using updated values for some (but not all)
points instead of the previous iteration's values (where multiplied
by 1-hcAlpha).

### DIFF
--- a/src/components/TweakBrush.cpp
+++ b/src/components/TweakBrush.cpp
@@ -511,7 +511,6 @@ TB_SmoothMask::~TB_SmoothMask() {
 
 void TB_SmoothMask::lapFilter(mesh* refmesh, const int* points, int nPoints, std::unordered_map<int, Vector3>& wv) {
 	std::unordered_map<int, Vector3>::iterator mi;
-	Vector3 d;
 	int adjPoints[1000];
 	int c = 0;
 	int a;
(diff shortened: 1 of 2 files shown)


sts1skj:
@ousnius: I could also templatize lapFilter and hclapFilter, if you like, so that there's one copy of each instead of three.

ousnius:
Smoothing seems to work way better now, thanks! If the templating doesn't get too complicated/confusing, go ahead. Otherwise I'd prefer more readable code to templates.

PS: There's some warnings in MSVC++ about unused variables (int a) and multiple declarations for int i.

src/components/TweakBrush.cpp
```diff
@@ -511,28 +511,17 @@ TB_SmoothMask::~TB_SmoothMask() {
 
 void TB_SmoothMask::lapFilter(mesh* refmesh, const int* points, int nPoints, std::unordered_map<int, Vector3>& wv) {
 	std::unordered_map<int, Vector3>::iterator mi;
-	Vector3 d;
 	int adjPoints[1000];
 	int c = 0;
 	int a;
```
ousnius: Unused variable, there's multiple like these.

sts1skj:
Hopefully I've fixed those warnings from MSVC.

sts1skj:
I'm giving up on templatizing lapFilter and hclapFilter.  I don't see how to do it nicely.

PR #238, opened 2020-01-29
Fixed three bugs in move brush

sts1skj:
1.  When the move brush overlapped the centerline and x-mirroring was
on, vertices on the other side of the centerline and inside the brush
radius were not mirror-moving.  I added code to TB_Move::brushAction
to check whether each vertex is closer to the origin or mirror-origin.

2.  Rarely, when using the move brush, pointStartState is set
but pointEndState is not for some vertices.  This resulted in
garbage being sent to the slider diff update when slider editing
was on.  I couldn't figure out how this was happening (it should
be impossible), but I prevented the transfer of garbage by adding a
check to ActiveShapesUpdated.

3.  There was a sign bug in GLSurface::CollidePlane.  This bug
caused the move brush to often move points less than it should,
and sometimes in the opposite direction.  I fixed the sign bug.
Of the many places from which CollidePlane is called, the sign bug
was already compensated for in all places except UpdateBrushStroke
in the TBT_MOVE code, which was why it was leading to problems only
for the move brush.  I removed the sign-error compensation from all
of the other places from which CollidePlane is called.

### DIFF
--- a/src/components/TweakBrush.cpp
+++ b/src/components/TweakBrush.cpp
@@ -850,6 +850,8 @@ void TB_Move::brushAction(mesh* m, TweakPickInfo& pickInfo, const int*, int, Twe
 		for (int p = 0; p < meshCache->nCachedPointsM; p++) {
 			int i = meshCache->cachedPointsM[p];
 			vs = startState[i];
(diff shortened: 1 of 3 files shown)


ousnius:
Thanks for this as well!

PR #239, opened 2020-01-29
Addweightbalance1

sts1skj:
Added weight-balancing to weight brushes
    
This is the first of a sequence of changes to add automatic weight balancing to Outfit Studio.  This change just affects the three weight brushes.  With this change, the weights for the bones on a vertex will be maintained to have more or less the same sum as they did before.
    
A lot of the changes in this commit are designed to support the next changes, which will add balance-bone selection to the GUI.

Please let me know if you'd prefer that I do one huge pull request for all of the weight-balancing stuff rather than doing it incrementally like this.  I thought you would prefer finer control and feedback as I add features one by one.

### DIFF
--- a/src/components/TweakBrush.cpp
+++ b/src/components/TweakBrush.cpp
@@ -1062,8 +1062,11 @@ void TB_Weight::brushAction(mesh* refmesh, TweakPickInfo& pickInfo, const int* p
 			if (bw.find(i) != bw.end())
 				totW += bw[i].endVal;
 		}
-		if (totW - 1.0 > -EPSILON) totW = 1.0;
 		float sw = ts.boneWeights[0].weights[i].endVal;
+		bool solo = false;
+		if (totW - sw <= EPSILON) solo = true;
+		if (solo) totW = 1.0;
+		if (totW - 1.0 > -EPSILON) totW = 1.0;
 		float ew = bFixedWeight ? strength * 10.0f - sw : strength;
 		ew *= getFalloff(pickInfo.origin.DistanceTo(refmesh->verts[i]));
 		ew *= 1.0f - refmesh->vcolors[i].x;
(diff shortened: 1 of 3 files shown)


ousnius:
@sts1skj Separate pull requests where possible are preferred so I can more easily see and understand the changes.

This seems fine but I'd like a way to turn this live normalization off. There are some editing cases where you don't want it to normalize before you're done doing your thing. I assume the way it was implemented now the old logic of just adding the paint isn't there anymore.

When painting everything to one bone (100% red), this happens to the other weights that previously affected the area (they don't fully become zero).
image

sts1skj:
I've tweaked the numerics so hopefully the other weights will go to zero more consistently.  I've also fixed a bug that was preventing weight from increasing when all other weights were zero.

As for wanting the option to not normalize, here was my plan for the user interface:

I would add an icon next to each bone (similar to the visibility icon for shapes) that would indicate "balance bone" (on) or "not balance bone" (off).  Bones would default to off.  If no bones are balance bones, all non-selected bones are used as balance bones (are included in the normalization).  If only a few non-selected bones are balance bones, weight would only be subtracted from those bones' weights.

In this scheme, if you want to change the weight of a bone without it being normalized, pick one or more balance bones that are already zero for that area.  Then no bone weights would change other than the selected bone's.

When I originally suggested this feature, I used the term "complementary bone".  I like the term "balance bone" better now.  If you can think of a better term, please tell me it.

ousnius:
I would probably just actually use "Normalized Bones", because that's what it does. Complementary and balance aren't really explaining what it actually does, so one can just use the real technical term (normalization) that at least people familiar with 3D modelling or some math might be familiar with.

Zeroing out works well now. 👍

sts1skj:
@ousnius Are you waiting for me to do something?

ousnius:
@sts1skj Sorry, I was busy. I will merge but before a future release of a new version I will have to go through and test everything thoroughly again.

PR #242, opened 2020-02-01
Consolidated multiple TweakUndo instances

sts1skj:
There were multiple instances of the undo history (class name TweakUndo),
one main one and one for each slider.  I consolidated them into one, the
main one.

I also made sure undo and redo work correctly across entering and leaving
slider edit.


ousnius:
@sts1skj After going from a regular undo to a slider undo, slider edit becomes activated but doesn't show one of the edits it should be showing. Redo doesn't have the problem.

The video is a bit confusing to follow as you can't see what was brushed before and what keys I'm hitting, but just try doing a normal edit after you did a slider edit and then undoing/redoing that and watch closely.

[video.zip](https://github.com/ousnius/BodySlide-and-Outfit-Studio/files/4143096/video.zip)

sts1skj:
@ousnius:
After constweakundo2, next I plan to separate the undo code from TweakBrush.{h,cpp} into separate files and rename some of the data structures:

UndoHistory.{h,cpp}:
TweakUndo -> UndoHistory

UndoState.h:
TweakStateHolder -> UndoStateProject
TweakState -> UndoStateShape
TweakBoneWeights -> UndoStateBoneWeights
TweakVertexWeight -> UndoStateVertexWeight

What do you think?  Would you like to make any changes to my suggested names before I start the work?

ousnius:
Names are fine. But if you plan to add any new additional features to undo/redo (like undoing project or shape changes), consult me beforehand so I can see if it's worth doing or if it should be done.

sts1skj:
I've been unable to reproduce the bug you reported, where undoing didn't work right for a sequence consisting a slider edit followed by a non-slider edit.  After undoing both, you should be in slider edit with no change showing; that's what happens for me.  Can you be more precise?

I've analyzed your video and tried my best to figure out what's going on.  Here's my best guess:

At the beginning of the video, your undo history has three items:
C1.  non-slider bulge of upper belly.
C2.  slider bulge of lower belly.
C3.  non-slider bulge of side belly.
You start in slider-edit mode with C1 and C2 applied but not C3.  Then you hit redo, taking you out of slider-edit mode, hiding C2, and applying C3.  Then you hit undo, unapplying C3.  Then you hit undo, taking you into slider mode and unapplying C2.  Then you hit undo, taking you out of slider mode and unapplying C1.  Then you hit redo, applying C1.  Then you hit redo, taking you into slider mode and applying C2.

Is my interpretation of the video correct?  If it is, is this not the correct undo-redo behavior?

Perhaps mode changes and model changes should not be consolidated in the undo history, so that switching into and out of slider-edit mode counts as a separate undo item.

ousnius:
Instead of rewatching my video and your analysis, I'll just go ahead and say yes to not having them consolidated. That will surely solve the issue I was having and also make it more clear to the user.

sts1skj:
I've changed the code so that mode changes are a separate undo/redo step, as you requested.  For me, this makes undo/redo less intuitive.  What do you think?

ousnius:
The bug I had I was talking about above is fixed with this.

To me it makes sense, as each thing you do including activating edit mode is part of the history. But on the other hand enabling/disabling edit mode doesn't actually change anything.

If we are going to revert it, then the bug needs to be fixed though.

Basically, after I undid a base change that leads into edit mode being automatically activated, the slider shape (at 100%) shows for one frame and then the base shape (slider at 0%) is shown instead. For redo this issue doesn't occur.

sts1skj:
I'm fine with it as it is now.

PR #243, opened 2020-02-02
Renamed and split undo classes from TweakBrush

sts1skj:
I renamed the following classes:
TweakVertexWeight -> UndoStateVertexWeight
TweakBoneWeights -> UndoStateBoneWeights
TweakState -> UndoStateShape
TweakStateHolder -> UndoStateProject
TweakUndo -> UndoHistory

UndoHistory was moved into UndoHistory.h and UndoHistory.cpp.  The rest were moved into UndoState.h.

I also renamed all of the functions in UndoHistory and most instances of variables of any of these types.

Keeping track of TweakStroke objects was moved from TweakUndo into wxGLPanel.  Now there is only ever one TweakStroke object.  I left the constant re-allocation of this object in place; perhaps it should just be re-initialized instead of re-allocated.

The state-applying functionality of TweakStateHolder::RestoreStartState, TweakStateHolder::RestoreEndState, TweakUndo::backStroke, TweakUndo::forwardStroke, wxGLPanel::UndoStroke, and wxGLPanel::RedoStroke was consolidated into one function, wxGLPanel::ApplyUndoState.  I tried not to change what this code does in any way, but the process was fairly messy, so I could easily have made a mistake.  Hopefully the result is much simpler and more maintainable.  Plus, it may be convenient to call ApplyUndoState from other places in the future.

Probably eventually UndoStateProject should be made polymorphic, but I feel Outfit Studio's project data model should be cleaned up first, probably including major changes to UndoStateProject since its purpose is to record changes to the project data.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ set(OSsources
 	${commonsources}
 	src/components/RefTemplates.cpp
 	src/components/TweakBrush.cpp
+	src/components/UndoHistory.cpp
 	src/files/FBXWrangler.cpp
 	src/program/EditUV.cpp
 	src/program/FBXImportDialog.cpp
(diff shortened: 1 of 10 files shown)


ousnius:
Fixed a compile error in MSVC++. First tests seem fine, I will merge and test more later on.

PR #245, opened 2020-02-04
Made weight copy auto-normalize

sts1skj:
I changed OutfitProject::CopyBoneWeights and the functions that call it
so that bone weights are normalized after being copied.  This affects
the commands "Copy Bone Weights" and "Copy Selected Weights" in the UI.

### DIFF
--- a/src/components/TweakBrush.cpp
+++ b/src/components/TweakBrush.cpp
@@ -910,7 +910,7 @@ TB_Weight::~TB_Weight() {
 
 void TB_Weight::brushAction(mesh* refmesh, TweakPickInfo& pickInfo, const int* points, int nPoints, UndoStateShape &uss) {
 	BoneWeightAutoNormalizer nzer;
-	nzer.SetUp(&uss, animInfo, refmesh->shapeName, boneNames, lockedBoneNames, bSpreadWeight);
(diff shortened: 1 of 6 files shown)


sts1skj:
@ousnius A couple of questions:
1.  May I make it so that the undo history isn't cleared when the selected bone is changed?
2.  I made "Copy Bone Weights" ignore the normalize-bone set specified by the user.  Should it instead limit the copy to the normalize bones?

ousnius:
@sts1skj 
1. If you can make sure selecting and editing different bones + weight copy can all be undone/redone properly without confusing one another, of course.
2. I think ignoring it is better. It would be way too confusing and potentially make a mess to copy weights with locked normalization to specific bones only.

ousnius:
Can you explain the way you moved the AddBoneAndXForm out of the copy function? Is the behavior for adding bones still the same as before?

sts1skj:
On AddBoneAndXForm: I noticed that the code of this function was duplicated in CopyBoneWeights and TransferSelectedWeights.  I extracted it and called it from both places.  Then I realized it should probably also be called from OutfitStudioFrame::ActiveShapesUpdated, so I added a third call there.  Then I realized the call from CopyBoneWeights was no longer needed.

Hopefully I didn't change the behavior for adding bones in any way.

PR #246, opened 2020-02-05
Made selecting a bone not clear the undo history

sts1skj:
I made it so that the undo history is not cleared when a bone is
selected.  I also tried to make sure that bone GUI elements update
at the right times.

### DIFF
--- a/src/program/OutfitStudio.cpp
+++ b/src/program/OutfitStudio.cpp
@@ -9007,13 +9007,9 @@ void wxGLPanel::ApplyUndoState(UndoStateProject *usp, bool bUndo) {
 			m->QueueUpdate(mesh::UpdateType::VertexColors);
 		}
 		os->ActiveShapesUpdated(usp, bUndo);
-		wxArrayTreeItemIds selItems;
-		os->outfitBones->GetSelections(selItems);
-		if (selItems.size() > 0) {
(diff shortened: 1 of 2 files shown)


src/program/OutfitStudio.cpp
```diff
@@ -2003,6 +2002,15 @@ void OutfitStudioFrame::GetNormalizeBones(std::vector<std::string> *normBones, s
 	}
 }
 
+std::vector<std::string> OutfitStudioFrame::GetSelectedBones() {
+	std::vector<std::string> boneList;
+	wxArrayTreeItemIds selItems;
+	outfitBones->GetSelections(selItems);
+	for (int i = 0; i < selItems.size(); i++)
+		boneList.push_back(outfitBones->GetItemText(selItems[i]));
```
ousnius: Doesn't compile for me on my wxWidgets and MSVC++ (pushing wxString into std::string vector).

sts1skj:
Hopefully commit 54b4005 fixes that compilation problem for MSVC++.

ousnius:
@sts1skj That wasn't the right line. You can revert that, line 2010 is the one that doesn't compile.

sts1skj:
Oops.  Hopefully I got it right this time.

PR #247, opened 2020-02-05
Clean: UndoStateShape indexing meshptr to string

sts1skj:
I changed the indexing of UndoStateShape in UndoStateProject (member
variable usss) to use a string shape name instead of a mesh pointer,
which should make it a bit more robust.

### DIFF
--- a/src/program/OutfitStudio.cpp
+++ b/src/program/OutfitStudio.cpp
@@ -8935,7 +8935,6 @@ void wxGLPanel::EndPivotPosition() {
 
 	std::vector<mesh*> refMeshes = activeStroke->GetRefMeshes();
 	if (refMeshes.size() > 0) {
-		mesh* pivotHitMesh = refMeshes[0];
 		UndoStateShape &uss = activeStroke->usp.usss[0];
 		if (uss.pointStartState.size() > 0 && uss.pointEndState.size() > 0) {
 			Vector3& pivotStartStatePos = uss.pointStartState[0];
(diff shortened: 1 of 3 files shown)


sts1skj:
After this, I was thinking of adding x-mirror bone functionality.  The main question I have is where to put the new user-interface element.  I was thinking of making a drop-down menu with the options "None", "Automatic", and all of the project bones.  Maybe it would be best to put the drop-down menu in the brush settings pane, maybe somewhere else.

ousnius:
@sts1skj I would do it like this:
image

The choice would need to be filled with the bone names everytime a bone is added, removed or the bone tree is reloaded.

The checkbox and choice need to be set to hidden/visible where the bone scale slider is set to hidden/visible as well.

Edit: Actually the "None" choice isn't needed, that would be disabling the checkbox.

src/program/OutfitStudio.cpp
```diff
@@ -8932,15 +8936,12 @@ void wxGLPanel::EndPivotPosition() {
 	std::vector<mesh*> refMeshes = activeStroke->GetRefMeshes();
 	if (refMeshes.size() > 0) {
 		mesh* pivotHitMesh = refMeshes[0];
```
ousnius: "pivotHitMesh" (line 8938) is now unused. Can the line be removed?

sts1skj: Yes.  Shall I do it?

PR #248, opened 2020-02-09
Added x-mirror bone feature

sts1skj:
Added a GUI element that appears when the bones tab is selected that allows
the user to select an x-mirror bone.  Changes by any of the three brushes
(weight, unweight, smooth weight) are x-mirrored onto the x-mirror bone.
There is also an "Auto" selection for the x-mirror bone.

### DIFF
--- a/src/program/OutfitStudio.cpp
+++ b/src/program/OutfitStudio.cpp
@@ -2061,7 +2061,7 @@ std::string OutfitStudioFrame::GetXMirrorBone() {
 	else if (xMChoice == 1)
 		return autoXMirrorBone;
 	else
-		return cXMirrorBone->GetString(xMChoice);
+		return cXMirrorBone->GetString(xMChoice).ToStdString();
 }
 
 void OutfitStudioFrame::SelectShape(const std::string& shapeName) {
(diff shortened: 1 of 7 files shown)


src/program/OutfitStudio.cpp
```diff
@@ -869,7 +869,11 @@ OutfitStudioFrame::OutfitStudioFrame(const wxPoint& pos, const wxSize& size) {
 		lightDirectional2Slider->SetValue(directional2);
 	}
 
+	wxPanel* m_panel6 = (wxPanel*)FindWindowByName("m_panel6");
+	m_panel6->SetBackgroundColour(wxColour(112, 112, 112));
```
ousnius: Why is this required? Is there a way to do it more cleanly (in the XRC maybe)?

sts1skj: I tried setting m_panel6's background color in the xrc, and wxWidgets for gtk ignored it.  Without the background color being set, all of the text labels in the panel are white on an off-white background for me.  (That's "De-/Select Sliders", "Fixed Weight Brush", "X-Mirror Bone", and "Preview Scaling".)

sts1skj: The parent panel of m_panel6 is bottomSplitPanel.  I tried setting bottomSplitPanel's background color instead to see if the color propagates to the child panel, and it doesn't.  The panel that needs its color set is m_panel6.

ousnius:
I gave it a quick spin and it seems to work very well so far, thank you!

PR #250, opened 2020-02-20
Cleaned up the coordinate-system transform code.

sts1skj:
Fixed scaling bug in MatTransform::ToMatrix.

Moved the implementation of MatTransform::ToEulerDegrees into Matrix3.
Made a radians version, ToEulerAngles.  Used ToEulerAngles
instead of ToEulerDegrees in Automorph::MeshFromNifShape and
GLSurface::AddMeshFromNif so that those functions don't have to
convert from degrees to radians when passing the values to glm::yawPitchRoll.

Added Matrix3::operator*(const Vector3 &).

Fixed bug in Matrix3::operator*(const Matrix3 &) where the identity
matrix was used instead of this matrix.

Fixed bug in Matrix3::operator*=(const Matrix3 &) where the two matrices
were swapped in the calculation.

Added Vector3::operator[].

Added Vector3 operator*(float, const Vector3 &).

Added Matrix3 Determinant, Invert, and Inverse functions.

Added MatTransform functions: ApplyTransform, InverseTransform, and
ComposeTransforms.  Used these functions in many places to simplify code.

Added NiAVObject functions GetTransformToParent and
SetTransformToParent and made everything except IO use these functions
rather than accessing "transform" directly.  I did this purely for
clarity, not for data hiding.

Simplified NifFile::GetAbsoluteNodeTransform.  Added a new name for it:
GetNodeTransformToGlobal.

Split functionality of NifFile::GetShapeBoneTransform into
GetShapeTransformFromGlobalToSkin and GetShapeTransformFromSkinToBone.
The old function now calls one of the two.  Same for SetShapeBoneTransform.

Added new name for NifFile::GetNodeTransform: GetNodeTransformToParent.
Same for SetNodeTransform.

Renamed AnimWeight::xform to xformSkinToBone.

Renamed AnimInfo::GetShapeBoneXForm to GetXFormSkinToBone.  Same for
SetShapeBoneXForm.

Replaced AnimBone's rot, trans, scale, localRot, and localTrans with
two MatTransforms: xformToGlobal and xformToParent.

Added MatTransform xformGlobalToSkin to AnimSkin.  This is set when
the Nif is read, but it's never written back out to the Nif.  Added a
line to OutfitProject::ResetTransforms to clear this transform, though
I can't see why calling ResetTransforms would ever be a good idea, so
I'm not sure if xformGlobalToSkin needs to be reset there.

AnimInfo::AddShapeBone is called from three places.  In all three
places, xformSkinToBone was being incorrectly set afterwards.  So I
added code to AddShapeBone to set xformSkinToBone to a good default
using the skeleton, and I eliminated the incorrect code from two of
the three callers (AddBoneRef and AddCustomBoneRef).  For the third
(AddShapeBoneAndXForm), the original code was supposed to copy the
xform from the base shape; I fixed it.

Eliminated AnimInfo::GetBoneXForm, which is no longer used.

Renamed AnimSkeleton::GetBoneTransform to GetBoneTransformToGlobal.

Made AnimInfo::WriteToNif use AnimWeight::xformSkinToBone in all cases
instead of recalculating it for one case.  The code for recalculating
it was incorrect, anyway (if the global-to-skin transform was messy).

Fixed probably harmless bug in AnimInfo::WriteToNif that was making
NifFile::SetShapeBoneWeights called if not FO4, when the proper
test was not BSShape.  (That is, I changed "if (!isFO4)" to "if
(!isBSShape)".)

Fixed likely bug in AnimInfo::CalcShapeSkinBounds where radius was
not being scaled.

Eliminated AnimSkeleton::GetSkinTransform, which was no longer being called.

Commented out lines in AnimSkeleton::GetBoneTransformToGlobal and
AnimInfo::WriteToNif that were setting scale to 1 for the skeleton's
bone-to-global transforms.  Normally, scale is already 1 for every
bone in the skeleton.  If they're not 1 for the skeleton, it's probably
for a good reason, so we shouldn't override that.

Cleaned up Automorph::MeshFromNifShape and GLSurface::AddMeshFromNif,
particularly for non-skinned meshes.  The transform for non-skinned
meshes is probably still wrong, but it should be closer to being right.

testing:
Loading CBBE_Body.nif and writing it back out: transforms unchanged.

Loading TypeGNoir_1.nif and writing it back out: transforms unchanged,
unlike before.

Add a bone, give it some weight, and write: transforms look good.

Copy bone weights to new bone and write: transforms look good.

Add a bone to TypeGNoir_1.nif, give it some weight, and write: transforms
look good.

For TypeGNoir_1.nif, for shapes with scale!=1, bounding spheres don't
appear correctly in NifSkope.  I don't know if this is a bug in NifSkope
or bad data in TypeGNoir_1.nif.  When Outfit Studio writes out this file,
the bounding spheres are completely different, and NifSkope still doesn't
display them correctly.  I don't know if this is a bug in NifSkope or
Outfit Studio.

Weight painting doesn't work right for shapes with scale!=1 in Outfit
Studio.  My best guess is that this is because of brush size not scaling.
Other brushes have similar problems.

### DIFF
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -1785,8 +1785,6 @@ int NifFile::GetShapeBoneWeights(NiShape* shape, const int boneIndex, std::unord
 }
 
 bool NifFile::GetShapeTransformGlobalToSkin(NiShape* shape, MatTransform& outTransform) {
-	outTransform = MatTransform();
-
 	if (!shape)
 		return false;
 
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -435,14 +435,14 @@ bool NifFile::DeleteUnreferencedBlocks() {
 	return hadDeletions;
 }
 
-int NifFile::AddNode(const std::string& nodeName, const MatTransform& xform) {
+int NifFile::AddNode(const std::string& nodeName, const MatTransform& xformToParent) {
 	auto root = GetRootNode();
 	if (!root)
 		return 0xFFFFFFFF;
 
 	auto newNode = new NiNode();
 	newNode->SetName(nodeName);
-	newNode->transform = xform;
+	newNode->SetTransformToParent(xformToParent);
 
 	int newNodeId = hdr.AddBlock(newNode);
 	if (newNodeId != 0xFFFFFFFF)
@@ -774,8 +774,8 @@ NiShape* NifFile::CloneShape(NiShape* srcShape, const std::string& destShapeName
 				// Move existing node to non-root parent
 				auto oldParent = GetParentNode(node);
 				if (oldParent && oldParent != nodeParent && nodeParent != rootNode) {
-					MatTransform xform;
-					srcNif->GetNodeTransform(boneName, xform);
+					MatTransform xformToParent;
+					srcNif->GetNodeTransformToParent(boneName, xformToParent);
 
 					std::set<Ref*> childRefs;
 					oldParent->GetChildRefs(childRefs);
@@ -784,7 +784,7 @@ NiShape* NifFile::CloneShape(NiShape* srcShape, const std::string& destShapeName
 							ref->Clear();
 
 					nodeParent->GetChildren().AddBlockRef(boneID);
-					SetNodeTransform(boneName, xform);
+					SetNodeTransformToParent(boneName, xformToParent);
 				}
 			}
 
@@ -1041,7 +1041,7 @@ OptResult NifFile::OptimizeFor(OptOptions& options) {
 				bsOptShape->GetProperties() = shape->GetProperties();
 				bsOptShape->GetExtraData() = shape->GetExtraData();
 
-				bsOptShape->transform = shape->transform;
+				bsOptShape->SetTransformToParent(shape->GetTransformToParent());
 
 				bsOptShape->Create(vertices, &triangles, uvs, normals);
 				bsOptShape->flags = shape->flags;
@@ -1307,7 +1307,7 @@ OptResult NifFile::OptimizeFor(OptOptions& options) {
 				bsOptShape->GetProperties() = shape->GetProperties();
 				bsOptShape->GetExtraData() = shape->GetExtraData();
 
-				bsOptShape->transform = shape->transform;
+				bsOptShape->SetTransformToParent(shape->GetTransformToParent());
 				bsOptShape->flags = shape->flags;
 
 				// Move segments to new shape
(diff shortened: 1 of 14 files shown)


sts1skj:
Here are a few things I found that maybe need to be cleaned up:

1.  OutfitProject::ResetTransforms sets each shape's global-to-skin transform to the identity, but it doesn't update the vertex coordinates or each bone's skin-to-bone transform.  Perhaps those should be updated too.

2.  There's a coordinate-system transform that is hard coded in a lot of places that I would call global-to-render or render-to-global.  In the global-to-render direction, it divides the coordinates by 10, negates the x coordinate, and swaps the y and z coordinates (so it's a scale and a rotation).  I don't understand why vertices are sometimes kept in render coordinates and sometimes in global coordinates.  It seems like it would be simpler to keep everything in global coordinates and let OpenGL apply the global-to-render transform as part of its model-to-view matrix.  In particular, I'm not sure if GLSurface::AddMeshFromNif is applying the global-to-render transform correctly, especially for non-skinned meshes in which the shape's node-to-global transform has a non-identity rotation.

ousnius:
@sts1skj Before I merge, can you please do some more tests with the following NIF files?
- How they import/export with OS
- How they look in OS (transform rendering)
- Changes to transforms (shape, global, bone, nodes) after exporting or adding bones/painting/copying

[Transform Tests.zip](https://github.com/ousnius/BodySlide-and-Outfit-Studio/files/4241237/Transform.Tests.zip)

sts1skj:
When I load "Head (00013659).nif" in Outfit Studio, it shows the shape "FemaleHeadHighElf" separate from the other shapes; they don't line up.  This happens for both new and old versions of Outfit Studio (both after and before I did any transform cleanup).  Is this a pre-existing bug in OS?  Or is this correct behavior?  In NifSkope, the shapes do line up.

ousnius:
@sts1skj That's correct behavior for now as OS doesn't render bone transforms. And heads are a special case in the game anyway. In NifSkope you can disable skinning with the bone icon in the toolbar to compare positions without skinned transforms.

sts1skj:
I've done the following testing on the 13 nif files of "Transform Tests.zip":

1. Load in OS to see if everything looked okay.  The only one with a problem was "Head (00013659).nif", which was mentioned in a previous comment.

2. For skinned Skyrim nifs, add a body reference and copy a bone's weight for a bone with no weighting in the input nif.  Then delete the reference.

3. For skinned nifs (including Skyrim), add a skeleton bone and paint some weight for it.

4. Export to NIF.

5. Load the exported NIF in NifSkope.  Check appearance.  Check that pre-existing transforms are unchanged.  Check that new transforms have reasonable values.  For shapes with scale==1, check that the bounding spheres look reasonable.

The only problem I found is that often the exported NIF has different textures when displayed in NifSkope than the input NIF.  As this problem has nothing to do with transforms, it's not relevant for this test.

So testing appears to be successful.

sts1skj:
@ousnius Would you like me to make two "operator *" and use them instead of ComposeTransforms and ApplyTransforms?  Do you have any other name changes you'd like me to make?

ousnius:
@sts1skj Does the latest commit also work with the FO4 vanilla outfit I provided? (those have different transforms than the FO4 CBBE body)

ComposeTransforms and ApplyTransforms are fine for now, they're more self-explanatory.

sts1skj:
I just tested again with "FO4 Armor with bone transform (Children of Atom).nif".  I loaded it in OS, added the bone SPINE1, painted some weight for it, exported it to nif, loaded it in NifSkope, and checked the new transforms.  They were very close to those for Spine1_skin.  So yes, it does appear to work with the FO4 armor outfit you provided.

sts1skj:
I was thinking: maybe OutfitProject::AddShapeBoneAndXForm isn't necessary.  It calculates the skin-to-bone transform for a new bone by copying the base shape's skin-to-bone transform for that bone.  Because the two shapes have different skin CSs, the calculation is complicated.  But, in the end, the calculated skin-to-bone transform ought to be the same as that calculated in AnimInfo::AddShapeBone, except with more accumulated round-off error.

May I eliminate OutfitProject::AddShapeBoneAndXForm?

ousnius:
In the code (before changes), AddShapeBoneAndXForm was intended to add an already loaded transform (part of workAnim (or the reference shape for FO4)) to the target shape when transferring weights or adding a new bone to a shape via the weight brush.

This was done so that the reference (or the first loaded shape) would give its transforms for the bone to the target shape.

If you think this is now done automatically in a different way, then you can remove it, otherwise we'd have to check if it's still needed.

sts1skj:
I tested commit f986e8c against TypeGNoir_1.nif.

sts1skj:
@ousnius I'm frustrated with myself; I seem to be getting more and more careless and forgetful.  The patch I added yesterday (commit 66b38b2) fixes a bug that I should have caught before submitting this pull request.  The bug was easily demonstrated by changing some vertex bone weights and then checking the result in Skyrim.  I should have done that end-to-end test before submitting the pull request.

Can you think of anything I might have forgotten to check for this pull request?  I'm not sure I'm not forgetting something important.

By the way, I've gotten posing implemented (branch addpose).  I'll submit a pull request for that branch once you've accepted this cleantransforms pull request.

ousnius:
I haven't seen any issues so far double-checking the code and NIFs. If there's anything we missed it's going to have to be fixed after it occurs.

PR #251, opened 2020-02-29
Implemented posing

sts1skj:
-Added GUI elements that only appear when the bone tab is selected
that allow the user to turn posing on and off, select a bone to pose,
adjust the six DOFs of the pose, and reset one or all bone poses.
Posing is always off when the bone tab is not selected.  (Changes to
OutfitStudio.{xrc,h,cpp}.)

-Added functions RotVecToMat and RotMatToVec to
lib/NIF/utils/Object3d.{h,cpp}.  These change a rotation vector into
a matrix and vice versa.

-Fixed bugs in Vector2::operator!= and Vector3::operator!=.

-Added member variables poseRotVec, poseTranVec, and xformPoseToGlobal
to AnimBone for keeping track of the pose.  Added function
AnimBone::UpdatePoseTransform, which updates xformPoseToGlobal
when the pose changes.  Added code to AnimInfo::LoadFromNif and
AnimBone::LoadFromNif to initialize xformPoseToGlobal.

-Added code to OutfitProject::GetLiveVerts for applying the pose when
posing is on.  Added a bool "bPose" to OutfitProject that controls
whether GetLiveVerts applies the pose.

-Added code to OutfitStudioFrame::ActiveShapesUpdated to use
GetLiveVerts to update the mesh when updating weights with posing on.

-Made wxGLPanel::ApplyUndoState and wxGLPanel::EndBrushStroke not
call OutfitProject::ApplyBoneScale if the bone-scale slider position
is 0.  I did this because the code in ApplyBoneScale resets the mesh
vertices, preventing weighting changes from appearing.  The code in
ApplyBoneScale seems very strange to me, so I was afraid to touch it.

Things to be aware of:

-I put no effort into making the GUI pretty.  It needs editing
to match the style of the rest of Outfit Studio.  GUI styling is a
matter of personal taste, and Outfit Studio is not my project, so I
would prefer not to do this editing myself.

-Someone other than me should take a careful look at the new
code for updating the visuals in OutfitStudioFrame::ApplyPose and
OutfitStudioFrame::ActiveShapesUpdated to make sure it's right.

-I don't think I handled the interaction between posing and bone
scaling well.  I would rather get rid of bone scaling.

### DIFF
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -435,10 +435,11 @@ bool NifFile::DeleteUnreferencedBlocks() {
 	return hadDeletions;
 }
 
-int NifFile::AddNode(const std::string& nodeName, const MatTransform& xformToParent) {
-	auto root = GetRootNode();
-	if (!root)
-		return 0xFFFFFFFF;
+NiNode* NifFile::AddNode(const std::string& nodeName, const MatTransform& xformToParent, NiNode* parent) {
+	if (!parent)
+		parent = GetRootNode();
+	if (!parent)
+		return nullptr;
 
 	auto newNode = new NiNode();
 	newNode->SetName(nodeName);
@@ -446,9 +447,9 @@ int NifFile::AddNode(const std::string& nodeName, const MatTransform& xformToPar
 
 	int newNodeId = hdr.AddBlock(newNode);
 	if (newNodeId != 0xFFFFFFFF)
-		root->GetChildren().AddBlockRef(newNodeId);
+		parent->GetChildren().AddBlockRef(newNodeId);
 
-	return newNodeId;
+	return newNode;
 }
 
 void NifFile::DeleteNode(const std::string& nodeName) {
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -1785,8 +1785,6 @@ int NifFile::GetShapeBoneWeights(NiShape* shape, const int boneIndex, std::unord
 }
 
 bool NifFile::GetShapeTransformGlobalToSkin(NiShape* shape, MatTransform& outTransform) {
-	outTransform = MatTransform();
-
 	if (!shape)
 		return false;
 
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -435,14 +435,14 @@ bool NifFile::DeleteUnreferencedBlocks() {
 	return hadDeletions;
 }
 
-int NifFile::AddNode(const std::string& nodeName, const MatTransform& xform) {
+int NifFile::AddNode(const std::string& nodeName, const MatTransform& xformToParent) {
 	auto root = GetRootNode();
 	if (!root)
 		return 0xFFFFFFFF;
 
 	auto newNode = new NiNode();
 	newNode->SetName(nodeName);
-	newNode->transform = xform;
+	newNode->SetTransformToParent(xformToParent);
 
 	int newNodeId = hdr.AddBlock(newNode);
 	if (newNodeId != 0xFFFFFFFF)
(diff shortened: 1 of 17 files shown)


ousnius:
If you'd like, you can get rid of the bone scaling slider (and ApplyBoneScale) if you implement bone scaling as part of the poser.

sts1skj:
What is the purpose of bone scaling?

ousnius:
@sts1skj The bone scale slider was added to have the very limited ability to preview weight factors by seeing how the bone/weight would scale. In some cases the center of the bone isn't calculated properly with it, either.

Some bones are used for scaling, such as the belly. One can preview the scaling that way.

ousnius:
@sts1skj 
Question 1 and 2: Better not change these for now, minimal changes to the NIF are always preferred. You never know what NIFs need specific transforms to not change (FO4 hard-coded cloth physics, anything to do with HDT, etc...)

You need to add allowCustomTransforms, DisableCustomTransforms() and their functionality back. It's not used in Outfit Studio, but it's used in SSE NIF Optimizer or other code I have.

3000percentmanlyjaw:
@sts1skj 

adsfasdsfafdsfdfdfjonbby-1

sts1skj:
Bone scaling is currently applied _after_ posing; it seems to me that it ought to be applied _before_ posing.  The steps of the vertex processing chain could be better consolidated, with a single path that vertex data takes from nif to opengl.  Caching of intermediate results, with appropriate invalidation of caches, could improve performance.

sts1skj:
I've fixed all the bugs that I know about.  I've done a great deal of testing over the past week, so hopefully I've fixed everything.  Please let me know if you find any more.

sts1skj:
Here's my (probably incomplete, as I've grown quite forgetful) list of things that I found (while implementing, testing, and bugfixing the posing code) that could be improved:

-The vertex processing pipeline could be overhauled.  This could simplify the code and speed things up a little.  Right now, there are calls of ApplyBoneScale, GetLiveVerts, UpdateMeshVertices, and ClearBoneScale all over the place.  I haven't analyzed this problem thoroughly yet.  Doing this overhaul does not excite me, but I could do it.

-The global-to-render transform (scale by 10 and rotate) seems like it could be handled more straightforwardly by including it in opengl's model-to-view transform rather than applying it to the vertex data.

-The add-custom-bone interface needs the ability to specify the parent bone and the rotation.

-It would be nice if the add-custom-bone interface could also be used to edit a custom bone.

-Smoothing brushes only use vertices directly connected to the modified vertex (through triangle edges).  This makes them not work well if the connections are asymmetrical.  It would be nice if they used all vertices in a connected neighborhood, possibly with a weighting applied so farther vertices matter less than closer vertices.

-The move brush doesn't keep vertices constrained to the surface.  Much of the time (but not all of the time), when moving vertices, you want them to not move from the surface but to stay constrained to it.

-For the smooth-position brush, it would be nice if there were an option to restrict motion to be perpendicular to the surface.  It might also be nice if there were an option to restrict motion to be along the surface (but it might not).

-It would be nice if there were a tool for flipping non-welded interior triangle edges.

-It would be nice if there were a tool for splitting an edge with a new vertex, with triangles updated to maintain the surface.

-It would be nice if there were a tool for eliminating a vertex that's connected to no more than three other vertices, with triangles updated and not deleted, so that the surface is preserved.

ousnius:
@sts1skj Exporting a NIF or saving a project while posing is active saves the NIF using the posed shape, but I'd consider this a good side effect as it can be used to modify the shape that way. Still, people need to be careful not to hit CTRL+S while saving if they don't want to actually use it as a shape.

Posing also updates the mesh normals, which is fine I guess. "Lock Normals" still correctly locks them.

You could create issues for the custom bone interface and brush improvements. It's unlikely I'll look into brushes any time soon, but never know when I get around to coding other things.

I'd prefer to do the vertex processing overhaul myself if I deem it should be done.

ousnius:
Thanks for the work on posing, a few people have requested it. Using it for weighting can still be a little confusing as there's no IK or rotation limits that the real skeleton has, but it's still useful a lot.

PR #260, opened 2020-03-25
Simplified partition code

sts1skj:
The purpose of this batch of changes is to simplify the code for maintaining partitions, whose data is stored in the nif file in NiSkinPartition.  Whenever possible, I moved code for converting between different representations of the data into the NiSkinPartition class, and I generally tried to use the triangle-partition-indices (or "triParts") representation wherever it made sense outside of NiSkinPartition.  Hopefully the new functions are simple enough to be easily checked for correctness, and the old functions that have been simplified are a little easier to understand.

I broke this pull request into a sequence of five commits.  My hope was that each would be simple enough that it could be easily examined and understood.  For each of the five commits, I did some basic testing.


sts1skj:
This one was a lot of work.  While I'm reasonably sure that the core components of this update work correctly, I'm not at all sure I caught all of the places in NifFile.cpp where data needs to be updated.  In particular, I wouldn't be surprised if I missed at least one place where a line "skinPart->triParts.clear();" needs to be added.

sts1skj:
@ousnius For my next task, I was considering doing some simple cleanup of duplicated code in Skin.cpp and Geometry.cpp.  It seemed like I saw a lot of different notifyVerticesDelete functions that had essentially the same chunk of code at the beginning.  Perhaps the common code could be extracted into a function or two or three.

May I start doing this cleanup?  Would you like me to work on something else?  I also need to take a close look at the segmentation code at some point.  Or perhaps I should write some documentation (such as "how to fix pose-dependent rough spots in an outfit").

ousnius:
@sts1skj The current code had checks for if a mapped triangle really exists in the triangles array for OptimizeFor and I think one other place. The new code doesn't, so it could possibly cause crashes for dirty NIFs that don't have fully valid partitions. Can you look into it? Unfortunately I don't presently have such NIF.

Feel free to do the NIF stuff, but it's probably better to make a specific list beforehand so I can see what types of cleanup would be made.

For segmentation there is already a very old branch by Caliente, you can check it out.

sts1skj:
I modified GenerateTrueTrianglesFromMappedTriangles and GenerateMappedTrianglesFromTrueTrianglesAndVertexMap so that they check for triangles with bad vertex indices and delete them.  The rest of the new NiSkinPartition code appears to be robust to bad vertex indices.

NifFile::PrepareData already calls RemoveInvalidTris (which deletes shape triangles with bad vertex indices).  Should UpdateSkinPartitions also call RemoveInvalidTris?  UpdateSkinPartitions does assume that the shape triangles have valid vertex indices.

The code for matching NiSkinPartition triangles to shape triangles is in GenerateTriPartsFromTrueTriangles.  It's robust to partition triangles not being in the shape triangle list.  If a shape triangle is not in the partition triangles, its partition index in triParts is set to -1.

I tried to make all code robust to partition indices of -1 in triParts.  If SetShapePartitions finds any, it puts those triangles in a new partition.  GenerateTrueTrianglesFromTriParts discards triangles with a partition index of -1.  UpdateSkinPartitions ignores them and lets GenerateTrueTrianglesFromTriParts discard them.

Is there other checking for bad triangles that needs to be done?

ousnius:
@sts1skj The triangle checks are fine, I think. I did some tests and it seems deleting a partition isn't working right now.

Load CBBE SE, add a new partition, assign some triangles to it and then delete the partition and hit apply.

sts1skj:
I've made deleting partitions work correctly, I think.  Triangles of the deleted partition get assigned to some other partition, which I think was the original intention.

ousnius:
Basic tests seem to indicate that it works the same as previously now.

PR #264, opened 2020-03-25
Normalize checkbox

sts1skj:
I added a checkbox for turning automatic weight normalization on and off.


ousnius:
Unfortunately I had to turn live normalization off by default, too many people (strangers and friends) and myself think live normalization is too confusing to work with in most cases. Painting an unweighted mesh manually is very confusing when parts have weights you just painted and other parts don't, it will really feel like "random colors" (though mathematically it's not).

So being able to paint without normalization and THEN turn it on for fine tuning makes more sense in this case since people are so used to it by now with OS.

PR #268, opened 2020-03-27
Extracted common code from notifyVerticesDelete and others

sts1skj:
I created a new header file, lib/NIF/NifUtil.h, with some algorithm functions extracted from notifyVerticesDelete and other functions in lib/NIF, and called those new functions from a bunch of places, simplifying the code.


sts1skj:
@ousnius If you really want me to, I could write an algorithm for turning triangles into strips.  It would be complicated.  I have no idea where in the code it would be best to replace triangles with strips.  As far as I know, the whole point of using strips instead of triangles is to reduce data size: file size, main memory consumption, video memory consumption, and the time to transfer the data between these three (in the game, not in Outfit Studio).

I would prefer to investigate segmentation next, though.

ousnius:
@sts1skj We don't need triangles to strips, they're not used in SK and they don't even exist in SE/FO4

PR #271, opened 2020-03-28
Did new rules for node output

sts1skj:
-Added functions SetParentNode and GetNodes to NifFile.  GetNodes is just GetShapes with "Shape" replaced with "Node", so it's probably good.  SetParentNode's implementation is more questionable.

-Added a lot of code to AnimInfo::WriteToNif to handle all of the cases for transferring bone data into nodes in the nif.  The three new rules that I hopefully implemented are:

1. New standard bone nodes have the root node as parent.

2. Custom bones have their parent node set correctly, reparenting if necessary.

3. All known bones have their node-to-parent transform set.

Unfortunately, the code is very complex.  It likely has at least one bug.


sts1skj:
@ousnius I haven't done any testing yet.  Also, it would be good if you could double-check my new code in AnimInfo::WriteToNif and NifFile::SetParentNode.  This one was pretty messy.

sts1skj:
I've done some testing:

- Lenore appears to output correctly now.

- I checked my own dangly-merchant nif (custom bones) to make sure it output correctly.

- I added a standard bone to CBBE-body SSE, painted some weight, output, and checked.

- I loaded BloodSkirt03_1, output it, and checked the node tree and some transforms.

Can you think of anything else you'd like me to test?  I'm feeling a bit fatigued, so I'm not sure if I'm missing something.  I think this pull request is ready.

sts1skj:
I just discovered that the re-parenting code does not work correctly.

sts1skj:
The re-parenting code seems to work now.

ousnius:
I've been going through some common processes in OS and testing the FBX in combination with other things again and haven't noticed any issues. Will merge.

@sts1skj Feel free to drop by the Discord server any time, it's easier to chat about things there and you can (if ever needed) interact with other users.

sts1skj:
@ousnius I've joined your Discord server with username sts1lusl#3019 .  I haven't used Discord before.  I'll try to remember to keep it running.

PR #273, opened 2020-04-01
Segment simplification

sts1skj:
This PR changes the way segment triangle membership is represented in OutfitStudioFrame to the triParts representation.  The code for translating to and from the new representation is in BSSubIndexTriShape.

Incidentally, I changed the way segments are colored in the model view.  The coloring is now the same as for partitions.

### DIFF
--- a/lib/NIF/Geometry.cpp
+++ b/lib/NIF/Geometry.cpp
@@ -494,6 +494,22 @@ int NiShape::GetBoneID(NiHeader& hdr, const std::string& boneName) {
 	return 0xFFFFFFFF;
 }
 
+bool NiShape::ReorderTriangles(const std::vector<uint>& triInds) {
+	std::vector<Triangle> trisOrdered;
+	std::vector<Triangle> tris;
+	if (!GetTriangles(tris))
+		return false;
+	if (tris.size() != triInds.size())
+		return false;
+	for (uint id : triInds)
+		if (id < tris.size())
+			trisOrdered.push_back(tris[id]);
+	if (trisOrdered.size() != tris.size())
+		return false;
+	SetTriangles(trisOrdered);
+	return true;
+}
+
 
 BSTriShape::BSTriShape() {
 	vertexDesc.SetFlag(VF_VERTEX);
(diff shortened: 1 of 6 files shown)


ousnius:
@sts1skj There is still a crash here when adding more triangles to a sub segment and clicking apply. Tested on FO4 "CBBE Body Reduced", attached here.

std::vector<bool> selPartIDs is empty in ShowSegment() after applying.

[CBBEBodyReduced.zip](https://github.com/ousnius/BodySlide-and-Outfit-Studio/files/4432480/CBBEBodyReduced.zip)

sts1skj:
I was unable to reproduce the crash you mention.  Could you be more specific about the sequence of steps needed to reproduce?

When you write that selPartIDs is empty, do you mean a vector of size 0, or do you mean every element is false?  In what circumstance did this occur?

ousnius:
@sts1skj 1.) Open Outfit Studio in FO4 mode (built with MSVC and with a debugger)
2.) Load CBBEBodyReduced.nif
3.) Select the body shape
4.) Go to the Segments tab
5.) Select the last sub segment of the right arm: ``#3: Human 2 | R-Lo Arm``
6.) Add a few more triangles to it
7.) Click Apply, crash

The vector is empty (size 0).

Call stack:
- OutfitStudio Debug.exe!OutfitStudioFrame::ShowSegment(const wxTreeItemId & item, bool updateFromMask) line 4697
- OutfitStudio Debug.exe!OutfitStudioFrame::OnSegmentSelect(wxTreeEvent & event) line 4261

sts1skj:
@ousnius That's very strange.  It doesn't crash for me; selPartIDs is not empty; and it works correctly.

I noticed that the line numbers in your call stack don't match the line numbers in my copy of the file.

sts1skj:
If I do "git pull https://github.com/ousnius/BodySlide-and-Outfit-Studio.git dev", my usual command for getting changes, it says, "Already up to date."  If I do "git status", it says I have no modified files.  So I'm at a loss how your line numbers could be different from mine.

My line numbers for the two lines in your call stack are 4259 and 4699, 2 less and 2 greater than yours.

"git pull https://github.com/sts1skj/BodySlide-and-Outfit-Studio.git segsimp" also says I'm up to date.

ousnius:
@sts1skj I haven't pulled commit c68b8d5 which adds two lines.

sts1skj:
@ousnius The line number in ShowSegment would have been two less before that commit, making it 4697, which is a match.  But that change is at line 4614, which is after OnSegmentSelect, so it wouldn't affect the line numbers there.

ousnius:
After pulling (I didn't have any local changes before or after).

image

sts1skj:
That image shows just what I would expect from your description.  selPartIDs has size 0, resulting in a crash.  The size of selPartIDs is determined on line 4618, where it is set to CalcMaxSegPartID() + 1.  That means CalcMaxSegPartID() is returning -1.  It does so when there are no segment items in the segment tree, or if the data in the segment tree has somehow gotten corrupted (like if SegmentItemData is missing or SegmentItemData::partID is -1).  Normally, no segment or subsegment's partID can be set to anything lower than zero (see BSSubIndexTriShape::GetSegmentation), which would mean CalcMaxSegPartID would return at least 0 if there were any segments.

So it looks like the crash is being caused by a bug somewhere other than in the chunk of code in the debugger image you posted.  Is CalcMaxSegPartID really returning -1, as I surmise?  Is the segment tree empty?  Is some segment or subsegment's partID garbage?  Where is OnSegmentSelect being called from (probably CreateSegmentTree at line 4582)?

Maybe OnSegmentSelect is being called from someplace strange, like from the DeleteChildren call at the beginning of CreateSegmentTree (and only for the Windows version of wxWidgets).  If so, fixing this bug might mean adding "triParts.clear()" right before the call of DeleteChildren.

ousnius:
@sts1skj This fixed the crash, however now there are no segment colors being loaded/drawn until the first time of clicking "Apply".

sts1skj:
Commit 3815c1d should fix the problem with no segment colors being loaded.

ousnius:
Fixed and the edited NIF seems correct in NifSkope.
Thank you!

PR #275, opened 2020-04-02
Added copy-bone-weight option to also copy skin transform.

sts1skj:
-Added a checkbox with explanatory text to the copy-bone-weights dialog
(dlgCopyWeights) for copying the reference's global-to-skin transform,
and implemented it.

-Added a function OutfitStudioFrame::CalcCopySkinTransOption that
calculates whether it makes sense to show the option and what the default
value for the option should be.

-Added function AnimInfo::ChangeGlobalToSkinTransform that sets the
transform and updates all the skin-to-bone transforms.

-Added IsNearlyEqualTo functions to MatTransform, Matrix3, and
Vector3, and added a FloatsAreNearlyEqual function.  These are in
lib/NIF/utils/Object3d.h.

Testing: I loaded FO4 CBBEBody, exported it to obj, imported it from
obj, merged in the CBBEBody as a reference, and copied bone weights.
The option appeared and was checked.  I then saved and checked that the
two shapes had identical skin-to-bone transforms for Spine2 in NifSkope.

When the copy-transforms option is hidden in the dialog, it leaves
an empty space.  Perhaps there's some good way to make the dialog lay
itself out again, but I'm no expert in wxWidgets, so I left that bug
to someone else to fix.

### DIFF
--- a/lib/NIF/NifFile.cpp
+++ b/lib/NIF/NifFile.cpp
@@ -1820,6 +1820,33 @@ int NifFile::GetShapeBoneWeights(NiShape* shape, const int boneIndex, std::unord
 	return outWeights.size();
 }
 
+bool NifFile::CalcShapeTransformGlobalToSkin(NiShape* shape, MatTransform& outTransform) {
+	if (!shape)
+		return false;
+	if (GetShapeTransformGlobalToSkin(shape, outTransform))
+		return true;
+
+	// Now the nif doesn't have this transform, probably because it's
+	// a FO4 nif, so we will try to calculate it, since FO4 shapes almost
+	// always have a non-identity global-to-skin transform.
+	// Ideally, we'd use bone transforms from the skeleton file, but we
+	// don't have access to that here.
+	std::vector<std::string> bones;
+	GetShapeBoneList(shape, bones);
+	for (const std::string &bone : bones) {
+		MatTransform xformBoneToGlobal;
+		if (!GetNodeTransformToGlobal(bone, xformBoneToGlobal))
+			continue;
+		MatTransform xformSkinToBone;
+		if (!GetShapeTransformSkinToBone(shape, bone, xformSkinToBone))
+			continue;
+		// compose: skin -> bone -> global and invert
+		outTransform = xformBoneToGlobal.ComposeTransforms(xformSkinToBone).InverseTransform();
+		return true;
+	}
+	return false;
+}
+
 bool NifFile::GetShapeTransformGlobalToSkin(NiShape* shape, MatTransform& outTransform) {
 	if (!shape)
 		return false;
@@ -1861,7 +1888,7 @@ void NifFile::SetShapeTransformGlobalToSkin(NiShape* shape, const MatTransform& 
 bool NifFile::GetShapeTransformSkinToBone(NiShape* shape, const std::string& boneName, MatTransform& outTransform) {
 	if (!shape)
 		return false;
-	return GetShapeBoneTransform(shape, shape->GetBoneID(hdr, boneName), outTransform);
+	return GetShapeTransformSkinToBone(shape, shape->GetBoneID(hdr, boneName), outTransform);
 }
 
 bool NifFile::GetShapeTransformSkinToBone(NiShape* shape, const int boneIndex, MatTransform& outTransform) {
--- a/lib/NIF/NifFile.h
+++ b/lib/NIF/NifFile.h
@@ -180,6 +180,10 @@ public:
 	void SetShapeBoneIDList(NiShape* shape, std::vector<int>& inList);
 	int GetShapeBoneWeights(NiShape* shape, const int boneIndex, std::unordered_map<ushort, float>& outWeights);
 
+	// Looks up the shape's global-to-skin transform if it has it.
+	// Otherwise, try to calculate it using skin-to-bone and node-to-global
+	// transforms.  Returns false on failure.
+	bool CalcShapeTransformGlobalToSkin(NiShape* shape, MatTransform& outTransforms);
 	// Returns false if no such transform exists in the file, in which
 	// case outTransform will not be changed.  Note that, even if this
 	// function returns false, you can not assume that the global-to-skin
--- a/lib/NIF/utils/Object3d.h
+++ b/lib/NIF/utils/Object3d.h
@@ -21,6 +21,11 @@ typedef unsigned char byte;
 typedef unsigned short ushort;
 typedef unsigned int uint;
 
+inline bool FloatsAreNearlyEqual(float a, float b) {
+	float scale = std::max(std::max(std::fabs(a), std::fabs(b)), 1.0f);
+	return std::fabs(a - b) <= EPSILON * scale;
+}
+
 struct Vector2 {
 	float u;
 	float v;
@@ -267,6 +272,12 @@ struct Vector3 {
 			y = 0.0f;
 		if (fabs(z) < EPSILON)
 			z = 0.0f;
+	}
+
+	bool IsNearlyEqualTo(const Vector3& other) const {
+		return FloatsAreNearlyEqual(x, other.x) &&
+			FloatsAreNearlyEqual(y, other.y) &&
+			FloatsAreNearlyEqual(z, other.z);
 	}
 };
 
@@ -538,6 +549,12 @@ public:
 		p *= 180.0f / PI;
 		r *= 180.0f / PI;
 		return canRot;
+	}
+
+	bool IsNearlyEqualTo(const Matrix3 &other) const {
+		return rows[0].IsNearlyEqualTo(other.rows[0]) &&
+			rows[1].IsNearlyEqualTo(other.rows[1]) &&
+			rows[2].IsNearlyEqualTo(other.rows[2]);
 	}
 };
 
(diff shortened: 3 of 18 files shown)


ousnius:
@sts1skj 
> It showed up in the wrong place, higher.  (Is this because of some transform that's applied only when in FO4 mode?  I imported it in SSE mode, and it displayed fine.) 
> 
> It looks like Outfit Studio does some peculiar things when in FO4 mode.
> Am I right, that these are bugs in OS?  Or am I totally confused?

Since Outfit Studio currently does not render the (individual) bone transforms in its pipeline like NifSkope does, in FO4 target game mode, the view (camera only) is lowered by 120 (12 in GL units), since that's where all FO4 skinned meshes appear without bone transforms rendered.

Your imported OBJ matches in NifSkope only when skinning is enabled. This is an issue, as FO4 meshes have a lower precision for vertex positions (half floats instead of floats), so with the head area being located further away from the origin (0, 0, 0), it's possible for there to be visible distortion of meshes where there wasn't when the head area was near the origin.

So for FO4, the transform option when exporting OBJs causes something that's counter-productive and makes these meshes different from all other FO4 meshes upon import, which is too dangerous to have in the tool (meshes will be all over the place, some up some down).

I will have to test the first commit some later in the day.

sts1skj:
Should I undo commit 16319cd (option to transform to global on obj output)?  Is there an easy way to do that?

sts1skj:
Do you want me to do idea 2 (let user edit the global-to-skin transform, possibly in the shape properties editor)?

3000percentmanlyjaw:
If this was the last commit, `git reset --hard HEAD^` or simply via the [github history](https://help.github.com/en/desktop/contributing-to-projects/reverting-a-commit)

sts1skj:
I'm inclined to keep commit 16319cd but change the wording so it's clear that this option should usually not be chosen.

sts1skj:
Here's an idea (let's call it idea 6): add a second checkbox to the copy-weight dialog, this one titled "Transform geometry so its location in global coordinates does not change".  It would effectively reverse the transformation of idea 1 (commit 16319cd).  The code I wrote yesterday for calculating the default value for the "Copy global-to-skin transform" checkbox could instead determine the default value for the "Transform geometry" checkbox (so the "Copy global-to-skin transform" checkbox would always default to on).

The editor for the global-to-skin transform (idea 2) could also have a "Transform geometry so its location in global coordinates does not change" checkbox.

sts1skj:
Here's a dumb question: why does FO4 use 16-bit floating point, with only 11 bits of mantissa, when it could use 16-bit fixed point, with 16 bits of mantissa?  The results would be higher quality, and you wouldn't give up anything.  16-bit fixed point is also faster than floating point, and has been supported directly by OpenGL since the beginning.

ousnius:
@sts1skj I'm not sure what you're referring to, there is only one short float format for OpenGL/DirectX and that's the IEEE standard for 16-bit half floats. Also we can't do anything about the dev's decision. There is a full precision flag for FO4 that allows NIFs to store 32-bit floats again, but it doesn't work with segmentation (dismemberment) and possibly other things.

About idea 6, feel free to implement it, but I cannot guarantee it can be merged until I can deem the final result as user friendly/straight forward enough.

jonwd7:
> Here's a dumb question: why does FO4 use 16-bit floating point, with only 11 bits of mantissa, when it could use 16-bit fixed point, with 16 bits of mantissa? 
Then that wouldn't be floating point?  You can't fill 16 bits of a 16-bit fp with just mantissa.  You could have a 32-bit fp with 16 bits of mantissa I suppose but then it wouldn't be IEEE standard. 

> The results would be higher quality, and you wouldn't give up anything. 
Completely incorrect.  The exact opposite in fact.

> 16-bit fixed point is also faster than floating point, and has been supported directly by OpenGL since the beginning.

Could be correct, ignoring a few hundred MMX and SSE instructions on the CPU not to mention dedicated floating point units on the GPU.  But also, 16-bit integer vertices would make it equivalent to PS1 graphics... There's a reason that PS1 models were blocky and fingerless.

For half float, you get much greater precision near 0.0, and decreasing precision away from 0.0.  Your suggestion for fixed point would mean there can only be 0 or 1 (and 2,3,...) unless you have some sort of divisor, at which point then you're dividing every stored value by 1000 or something and turning it back into float anyway.  What exactly would that solve?

Half float gives you the opportunity of having vertices very close together without quantizing (snapping) like fixed point would, especially for values between 0.0 and 1.0.  Beyond 1.0 quantization becomes increasingly likely for finer details.

For larger meshes with fine details, that means you break them up and put the finer details centered around the origin, and push the mesh back into place with a transform.  The suggestion for fixed point would mean you can just never have those fine details period, unless you either divide all the values or increase the scale of your game by a few thousand so that fixed point vertices represent fine enough details.  Of course 16-bit can only go up to 65,535 which introduces a whole series of other problems. 

You could get even more precision out of half float by actually lowering the scale of your game.  I'm not sure that Bethesda's game scale was particularly great for half float, so that's one legitimate issue with half float that I can point out.  If they lowered their scale by 10 they'd probably get more precision in people-sized meshes without any downside. 

The biggest issue by far with half float is actually in SSE, where they converted all the game's UVs to half float and didn't even check for issues.  Hint:  There are issues.   But Bethesda's mistake doesn't mean half float is invalid, it means they should have checked the UV conversion.

sts1skj:
The latest commit 326ea4d should have everything, except proposed changes to the reset-transforms command.

ousnius:
@sts1skj The scale value in shape properties seems to do something weird to lighting/normals on CBBE SE (with and without applying the transform).

Note: CBBE SE has no mesh normals/tangents (it uses model space normal maps).

sts1skj:
@ousnius I suspect the scale-affecting-lighting bug was already present.  Letting the user change the scale at will just seems to demonstrate the bug.  The reason I think this is because of the example file "TypeGNoir_1.nif" you sent me a while ago.  I noticed when I first examined it that the skin was strangely shiny, as if the specular strength were set too high.  I ignored it, assuming that was the artist's choice.  Now that I can change the skin CS scale, though, I find that I can make the strange shininess go away by setting the skin CS scale to 1.1.  Setting the scale to something lower than 1 increases the shininess.

I can try to track down this bug, but it could take me a while, as I suspect I'll have to look far and wide through the code.  Of course, it's possible that it _is_ actually the result of something I broke in this PR.

sts1skj:
@ousnius I loaded SSE CBBEBody.nif in NifSkope, set the global-to-skin transform scale to 0.9, saved it, and loaded it in a version of Outfit Studio from the dev branch (before I started working on segsimp or skinTrans2).  It appeared glossy.  This seems to indicate that the glossiness/lighting problem is not the result of the changes in this PR.

sts1skj:
@ousnius I think I've found the scale-affecting-lighting bug.  In res/shaders/default.frag, there are two places where "normal" is calculated by multiplying a 3x3 submatrix of matModelView by a unit vector.  matModelView comes from the product of matView and matModel in GLShader::SetMatrixModelView.  matModel is constructed in GLSurface::AddMeshFromNif, which includes the inverse of the global-to-skin transform's scale in the matrix.  So the 3x3 submatrix used in the fragment shader includes a scale.  So "normal" is not a unit vector.  But lots of code in the fragment shader assumes that "normal" _is_ a unit vector.

If the global-to-skin scale is less than 1, "normal" ends up being longer than 1, which makes specular shininess calculations return a larger value than they should (NdotH is bigger than it should be), making the model shiny.  If the global-to-skin scale is less than 1, specular reflection is reduced, and diffuse lighting is also reduced (NdotL is less than it should be), so the model looks dimmer and less shiny.

I'm not sure what the best fix is, but here's a simple one:  add a line "normal = normalize(normal);" after each of the two lines where it's set using matModelView in default.frag.

matModel is also used in calculating "reflectedWS".  I don't know what it's used for (I'm far from being a GLSL expert), so I don't know whether "reflectedWS" is being adversely affected too.

I don't know whether FO4 might have a similar issue.

sts1skj:
@ousnius Perhaps it would be a good idea to make GLSurface::AddMeshFromNif get the correct global-to-skin transform for FO4 meshes; and to remove the camera offset for FO4.  There are two ways I can see that AddMeshFromNif could get the correct transform:

1. Instead of calling NifFile::GetShapeTransformGlobalToSkin, which returns the identity transform for FO4 meshes, grab the transform from AnimSkin.  I'm not sure how AddMeshFromNif would get access to AnimInfo and thus AnimSkin.

2. Move the code for calculating the transform for FO4 from AnimSkin::LoadFromNif into NifFile::GetShapeTransformGlobalToSkin.  This would be messy because NifFile doesn't have access to the reference skeleton AnimSkeleton, so it would have to rely on the transform data in NiNodes.

What do you think?  Should I go ahead and do it?  I'm leaning towards option 1 right now.

ousnius:
@sts1skj Method 1, but instead of passing AnimInfo to AddMeshFromNif, do the "m->matModel" setting calculations outside of the function for the returned mesh in a function in OutfitProject instead. This would cause a difference between BodySlide and Outfit Studio in rendering, though.

sts1skj:
@ousnius Hmm.  GLSurface::AddMeshFromNif is used by BodySlide as well as Outfit Studio.  As far as I can tell, BodySlide doesn't use AnimInfo or AnimSkeleton.

sts1skj:
@ousnius I wonder if Automorph::MeshFromNifShape should be updated as well, as it's the only other place NifFile::GetShapeTransformGlobalToSkin is called.  But I know nothing about this class and its purpose.

ousnius:
MeshFromNifShape sets data in a mesh whose vertices are then used for building a KD tree for copying weights and conforming shapes to sliders. If the vertices aren't transformed correctly, weights and sliders won't copy from the correct source location.

sts1skj:
Currently, for Skyrim SE, Outfit Studio's conform-to-sliders commands do not work for shapes that have different global-to-skin transforms, but the copy-weights commands do work.  If I update the code in Automorph::MeshFromNifShape, then copy-weights should work for FO4 as well as it does for SSE, but that won't fix the conform-to-sliders commands.

PR #276, opened 2020-04-05
Fix lighting problems for scaled shapes

sts1skj:
For skinned shapes with a global-to-skin transform with scale other
than 1, the lighting was wrong.  With scale < 1, the shape was too
shiny and bright.  With scale > 1, the shape was too unshiny and dim.
This was because normals were not being normalized in the non-FO4
fragment shader.


sts1skj:
This problem was discovered during testing of pull request 275.  There are messages about it in that thread.

PR #278, opened 2020-04-11
Implemented eliminate-vertex tool.

sts1skj:
The guts of the tool are in the new function OutfitProject::PrepareElimVertex, which figures out which vertices and triangles to delete and what new triangles should replace them.

- Added "CursorType" enum to GLSurface, with a member variable "cursorType" and get and set functions.  It affects the behavior of GLSurface::ShowCursor (allowing the brush circle to be hidden for the new tool).  Also added a HidePointCursor function.

- Fixed a bug in GLSurface::GetCursorVertex.

- Added new icons, a menu item, and a toolbar button for the new tool.

- The function wxGLPanel::SetActiveBrush had a parameter, brushID, which was a non-enumerated integer ID (so it was generally called with an integer literal argument).  The values seemed to be a subset of the values of the enum ToolID, so I changed its type from int to ToolID and replaced the integer literals with named constants from ToolID. I changed ToolID from a plain enum to an "enum class" and moved it out of OutfitStudioFrame.

- Split the bool "editMode" in wxGLPanel in two: editMode and brushMode.  editMode is now set when any edit tool is selected, whether or not it's a brush, while brushMode is set just for brushes.

- Added four member variables to wxGLPanel: hoverMeshName, hoverPoint, mouseDownMeshName, and mouseDownPoint.  These record the mesh and point that the mouse last moved over and where the mouse was last left-clicked.  They're used in the new eliminate-vertex code.

- Consolidated the bools bMaskPaint, bWeightPaint, and bColorPaint in wxGLPanel into the ToolID activeTool.

- Added a bool isVertexPicking to keep track of whether the mouse is down for eliminate-vertex (or potentially other future vertex-picking tools).

- Added a check to OutfitStudioFrame::SelectShape to do nothing if the selection isn't changing.  (It was clearing the undo history sometimes unnecessarily.)

- Made OutfitStudioFrame::OnTabButtonClick call SelectTool instead of bypassing it and calling SetActiveBrush directly.

- Implemented the new eliminate-vertex tool in OutfitStudio.cpp.  Most of the new functionality is in StartVertexPick, UpdateVertexPick, EndVertexPick, and ClickElimVertex.


sts1skj:
I'm not happy with my icon, but my artistic skill is too limited for me to figure out how to make it better.  I don't even know why I don't like it.

PR #279, opened 2020-04-13
Made ApplyTransformToShapeGeometry also apply to diffs.

sts1skj:
This affects the copy-bone-weights and shape-properties commands when
the user chooses to transform the geometry.

ScaleShape and RotateShape also should be updated at some point.  This
commit does not affect those functions.


sts1skj:
It turns out this change is undesirable: the diffs are always in the base shape's skin coordinate system, not the shape's own skin coordinate system.  So they should never be transformed.

PR #287, opened 2020-04-25
Fix split edge

sts1skj:
- FindNeighboring triangles wasn't checking for not-found triangles before trying to determine the opposite vertex.

- ClickSplitEdge was not checking for edge orientation mismatches when finding welded edges.  The code was moved into PrepareSplitEdge.

- PrepareSplitEdge was not handling uv diffs for welded edges properly.

Note that Split Edge still does not work correctly.  There's a flaw in kd_matcher that causes the welded vertex lists to be incomplete, which prevents Split Edge from working correctly if a vertex is welded to more than one other vertex.

### DIFF
--- a/lib/NIF/Geometry.cpp
+++ b/lib/NIF/Geometry.cpp
@@ -1134,7 +1134,7 @@ void BSTriShape::SetEyeData(const std::vector<float>& in) {
 		vertData[i].eyeData = in[i];
 }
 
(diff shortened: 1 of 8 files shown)


sts1skj:
@ousnius Since I touched the normals-calculating code, it would be good to check that normals are still being calculated correctly.  I did quite a bit of testing myself, but I doubt I covered every case.

ousnius (approved):

PR #290, opened 2020-04-30
Implemented copy-geometry (a.k.a. merge-shapes) command

### LINKED ISSUE #267: Merge two or more meshes into one
**What is it that you want to do that isn't currently possible?**
In outfit studio, there is an option to separate vertices. Which allows users to create two separate meshes. But there is no option to merge them back.
**Request the feature**
I suggest adding the "merge vertices" option when one or more meshes selected to Outfit Studio.

sts1skj:


sts1skj:
-More checks need to be added to OutfitProject::CheckMerge.

-Deleting partitions doesn't seem to work any more.  I'll look into that.

sts1skj:
The new copy-geometry dialog box doesn't resize properly as its contents change.  I'm not knowledgeable enough about wxWidgets to easily fix this.

In addition, the "OK" and "Cancel" buttons don't get the right size if the "OK" button is disabled.  The button's size should not change just because it's disabled, so I have no idea why it's doing this.  Perhaps it would be fixed if the first call of CheckCopyGeo somehow took place _after_ the dialog box is shown.

ousnius (approved):

PR #377, opened 2021-08-27
1 click 'Convert Body Reference' wizard (ie, CBBE to BHUNP) & 'Keep Zap Sliders' on loading a reference

daz001:
* Added the ability to keep zap sliders when loading a reference body
* ~~Added 'Slider->Bake Conversion Ref' menu item
      - This menu item can be used wehn converting between body references in order to exclude the need to 'export to nif with reference'. Instead, this will simply bake the bodyslide used from the conversion reference into the mesh and you can simply continue using the mesh.~~
* Added 'File->Convert Body Reference' menu item
      - This will allow you to 1 click convert between body references (ie. CBBE to BHUNP)
      - All settings that are set on the wizard dialog are remembered between sessions - so you only need to enter them once for all of the outfits you wish to convert
      
      How to use:
         1) Open your outfit project
         2) Open the wizard ('File->Convert Body Reference')
         3) Select the conversion reference (ie. Convert: CBBE SE to BHUNP)
         4) Select the new body reference (ie. BHUNP 3BBB Advanced ver 2)
         5) Enter any text that you would like removing from outfit name, slider set file, shape data folder and shape data file - including spaces (ie. 'CBBE, Physics, p, P, SSE')
         6) Enter any text that you would like prepending to outfit name, slider set file, shape data folder and shape data file - including spaces (ie. 'BHUNP, Physics, p, P, SSE')
         7) Click OK, wait for the process to complete and you can then instantly save the project (which will use the updated project name values) or adjust weights and then save.

### DIFF
--- a/res/xrc/OutfitStudio.xrc
+++ b/res/xrc/OutfitStudio.xrc
@@ -1661,6 +1661,11 @@
 				<label>Load Outfit...</label>
 				<help>Load a NIF file as the working outfit, replacing any current outfit objects.</help>
 			</object>
+      <object class="separator" />
+      <object class="wxMenuItem" name="fileConvBodyRef">
+        <label>Convert Body Reference...</label>
+        <help>Convert an outfits body reference to use a different reference</help>
+      </object>
 			<object class="separator" />
 			<object class="wxMenuItem" name="fileSave">
 				<label>Save Project\tCtrl+S</label>
--- a/res/xrc/OutfitStudio.xrc
+++ b/res/xrc/OutfitStudio.xrc
@@ -1662,10 +1662,10 @@
 				<help>Load a NIF file as the working outfit, replacing any current outfit objects.</help>
 			</object>
       <object class="separator" />
-      <object class="wxMenuItem" name="fileConvBodyRef">
-        <label>Convert Body Reference...</label>
-        <help>Convert an outfits body reference to use a different reference</help>
-      </object>
+			<object class="wxMenuItem" name="fileConvBodyRef">
+				<label>Convert Body Reference...</label>
+				<help>Convert an outfits body reference to use a different reference</help>
+			</object>
 			<object class="separator" />
 			<object class="wxMenuItem" name="fileSave">
 				<label>Save Project\tCtrl+S</label>
@@ -1964,10 +1964,10 @@
 				<help>Create a new Zap slider based on unmasked vertices</help>
 			</object>
 			<object class="separator" />
-      <object class="wxMenuItem" name="bakeNifInPlace">
-        <label>Bake Conversion Ref</label>
-        <help>Bake the conversion slider</help>
-      </object>
+			<object class="wxMenuItem" name="bakeNifInPlace">
+				<label>Bake Conversion Ref</label>
+				<help>Bake the conversion slider</help>
+			</object>
 			<object class="separator" />
 			<object class="wxMenuItem" name="sliderImportOSD">
 				<label>Import OSD...</label>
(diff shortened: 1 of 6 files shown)


daz001:
I expect there will be some recommendations - my knowledge over wxWidgets is obtained over the last day or so and the UpdateProgress related things were giving me a headache, so I silenced most of the sub process ones. 

Also, I have no idea why the OutfitStudio.cpp diff looks so crazy, I only added a few methods..

daz001:
FInally, fixed the formatting, not sure what to do about the progress bar stuff though - I just get exceptions if I try to leave subprogress things in. The idea was to show the progress for the entire wizard, but it doesn't seem to work as expected.

Though, functionality - everything is good and existing functionality with progress things are unaffected. It's just a bit messy

ousnius:
@daz001 I'm not at home for another week, so I can't check in detail just yet.

What's the difference between "Slider->Bake Conversion Ref" and "Slider -> Set Base Shape"?

Can you fix the .xrc indentation? The progress stuff will need to be fixed by someone (remove silent again) before merge.

daz001:
@ousnius Your suspicion was correct, SetBaseShape appears to have the same functionality, so I've removed BakeConversionRef functionality (which was a confusing name anyway) and replaced it with the use of Set Base Shape. 

It's strange that the conversion manuals i've looked at don't seem to mention the use of SetBaseShape, especially since it reduces the time needed to convert a reference dramatically over exporting a nif with reference. Possibly also a confusing name? Edit: nvm - just asked a modder and he knew about it :/

I've also fixed the xrc indentation and I'm going to look into the progress stuff next

daz001:
Ignore my last couple of commits - although they work perfectly, i've since come to realize the exact purpose of StartSubProgress and i'm going to be removing the min/max parameters

daz001:
@ousnius Ok - i've changed everything to the current way of implementing things (StartSubProgress calls from the highest method). Total progress for the method jumps between positions when calling methods like ConformShapes - this cant be solved easily unless passing min/max progress values to methods which kinda smells.

The current progress implementation of calculating progress from the stack makes this hard for progress to be correct after doing any refactoring for method reuse. I feel that updating the calculation to take into account the current progress position on the stack and interpolate appropriately would be more flexible in these scenarios. Currently, hard coded values eg. StartSubProgress(10,20) in method calls will mess up the progress of a whole running process (unless called from the highest level or the correct values are passed into the methods and used).

Going forward, I guess the options are - just get it looking good and pass the min/max progress values to methods that call StartSubProgress, leave as is (works fine, just the progress bar shows incorrect progress at certain points), or change how StartSubProgress works and progressValue is calculated. Unless i'm misunderstanding?

I have some other ideas for 1 click wizards also, but - its reusing existing code and placing it into a wizard type dialog which means combining multiple sets of 'progresses' that are hard coded into methods with specific progress values. I feel these tools are huge time savers when it comes to working with outfit studio and mod devs will appreciate time saved if we can do it nicely

daz001:
@ousnius I've updated the progress calculation to adjust for the last minimum value on the stack - so aslong as the highest level method sets the subprogress (which everything previously did), all lower level methods will adjust accordingly (min progress + scaled range), so now progress cant jump lower than the previous value when calling a lower level method. I also ensured that all of my lower level calls now have a higher level StartSubProgress with the correct values. The progress bar now updates correctly with my wizard and existing functionality seems unaffected and no smelly code was added !

Planned:
- Add checkboxes on the wizard for 'Conform With Defaults' and 'Copy Bones with Defaults' which will set the silent flag for the dialog popups.
- Add config save/loads for Keep Zap Sliders / Use Default (conform & copy) checkboxes and uncheck 'keep zap sliders' by default

daz001:
Improved the dialog by adding instructions, saving checkboxes to the config and adding skip popup options

Issue #386, opened 2021-12-02
Batch conflict resolution should be supported in the base UI


Iloei:
When I am working in bodyslide, I want to see what nif(s) I would be generating. And, when I have conflicting definitions for a nif, I want to preview the possibilities so that I can pick the one I like best. And, when I make this choice, it should be preserved for batch builds.

This is similar to Issue https://github.com/ousnius/BodySlide-and-Outfit-Studio/issues/350 (partial overlap, but this version is more limited in scope).

Iloei:
Pull request https://github.com/ousnius/BodySlide-and-Outfit-Studio/pull/387 would be a plausible implementation of this concept.

ousnius:
PR #387 merged

PR #387, opened 2021-12-02
View output conflicts and file paths in BodySlide

Iloei:
Exposes the bodyslide conflict management logic used in batch builds in the base UI.

Supports quick navigation between conflicting alternatives.

Implements Issue #386

This is a first draft -- if you do not like something about the code, please let me know and I will fix it.

(Note also that I have only tested this under VS2022 so there may be issues under VS2019)

### DIFF
--- a/res/xrc/BodySlide.xrc
+++ b/res/xrc/BodySlide.xrc
@@ -185,7 +185,7 @@
 								<option>0</option>
 								<flag>wxLEFT|wxALIGN_CENTER_VERTICAL</flag>
 								<border>5</border>
-								<object class="wxCheckBox" name="cbDefaultOutfit">
+								<object class="wxCheckBox" name="cbIsOutfitChoice">
 									<fg>#c8c8c8</fg>
 									<tooltip>Default outfit choice in Batch Build</tooltip>
 									<label></label>
--- a/res/xrc/BodySlide.xrc
+++ b/res/xrc/BodySlide.xrc
@@ -13,7 +13,7 @@
 				<flag>wxEXPAND</flag>
 				<border>5</border>
 				<object class="wxFlexGridSizer">
-					<rows>2</rows>
+					<rows>3</rows>
 					<cols>5</cols>
 					<vgap>0</vgap>
 					<hgap>0</hgap>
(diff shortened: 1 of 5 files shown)


ousnius:
@Iloei Thanks for contributing, I've taken a look at it, it seems like a nice addition.

So what this does on a technical level is:
- Show the output path of the current outfit in BodySlide
- White = no conflicts
- Cyan = Conflict but it's the current choice
- Red = Conflict and it's not the current choice

I don't have any issue with the draft as a whole, but I'd do some minor changes to the UI to make things clearer to the user about what things mean.

- Show the output file name (not just the path). In the case of Skyrim this is both _0 and _1.nif, for other games just one file.
- Move the "Output Path" to the bottom of BodySlide, since the checkbox/color only affects batch build and not regular builds since clicking "Build" always overwrites what you currently have installed. Maybe right between the "Preview" and "Build" Buttons.
- Right now you can only turn an outfit into the default selection (check the box), but you can't "remove" the selection (uncheck the box). Either make it possible to do so, because clicking a checkbox with nothing happening feels weird, or make the checkbox greyed out if it's checked. I'd prefer the former.
- Possibly more error checking when calling some functions, but I'd have to go through those and check. It might be fine.

Let me know if you don't want to do any of those yourself, then I can work on them instead.

Iloei:
1. Adding the full output file names makes sense to me. I will draft something up.
2.  I think I have figured out how to do this, using the wxGridBagSizer.
3. Removing the default selection when deselecting the checkbox should be doable. I'll put something together for that.
4. Thank you.

Iloei:
I have implemented the functional aspects of your request (and merged your dev changes).

ousnius:
@Iloei Thank you. I've implemented some more tweaks for better UI clarity/UX. Do those look fine to you as well?

Iloei:
That should be fine, I imagine.

(I like that now the initial display is cleaner. My only worry would be that the abbreviation of the file name might be too much for the people who are just starting... (but I think that people who are just starting would also be having other bigger problems and by the time they get those issues sorted that they would be fine with the file name abbreviation mechanism))

I also like that you explicitly call attention to the right click mechanism. Expecting people to "just know" to use the context menu was me asking too much of the user.

Thank you.

ousnius:
@Iloei I'm not so much worried about the file name abbreviation `(and _1.nif)` but more so about the checkbox and path/color being there in general. A new user might not understand what the checkbox does or what conflict means when they first see it, so it makes the UI unnecessarily more busy for them.

But I couldn't really think of another way or place to put it or make it more clearer. There's a wxCollapsiblePane where one could hide it under, but in wxWidgets it's really ugly and takes up too much space (and you can't change its background color). So this has to do for now.

ousnius (approved):

Iloei:
@ousnius If it becomes a problem, that part of the display could be left blank by default and such that it only appears if enabled in settings.

That would still be less than ideal, but should be relatively easy to implement.

That said, people with the technical ability to run and use bodyslide are probably capable of understanding the concept of .nif files (and this is a concept which many users of bodyslide are already familiar with).

PR #389, opened 2021-12-05
Update conflictLabel tooltip to be informative

Iloei:
Since the "Output Path" label has been removed, and the previous text of this tooltip is now redundant.

Also, only show info message about conflicts when conflicts exist.


ousnius:
Both makes sense, thanks.

ousnius (approved):

PR #402, opened 2022-01-17
Clamp max alpha

daz001:


daz001:
contemplating also added warnings for vertex_color / vertex alpha not being enabled (depending on tool selected)

ousnius (approved):

PR #403, opened 2022-01-17
Add wizard to convert/replace references

daz001:
Has the ability to 1 click convert a reference body in an outfit project (ie. CBBE to BHUNP) or replace an existing body (select 'None' for the conversion). Though, you may wish to do additional weight painting after the conversion is complete (depending on the outfit)

Includes an automated way to rename the project (which can be tedious when doing many outfits) and various other settings that may be useful

Tried my best to separate the dialog from the main code, though as it calls existing methods to automate the process, some methods required minor refactoring

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,7 @@ set(OSsources
 	src/program/OutfitProject.cpp
 	src/program/OutfitStudio.cpp
 	src/program/ShapeProperties.cpp
+	src/program/ConvertBodyReferenceDialog.cpp
 	)
 set(BSsources
 	${commonsources}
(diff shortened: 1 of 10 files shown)


ousnius:
@daz001 The filter for the conversion and reference choice is a good idea, but I'm unsure if it should be implemented.

There might be a lot of conversion references on the web (or personal ones) that don't follow the "Convert:" or "Sliders:" naming convention. What you could try doing instead, is sorting them to the start of the choice, instead of removing the others.

I should be able to merge this afterwards.

ousnius (dismissed):

ousnius (approved):

PR #407, opened 2022-01-21
Slider merge issue when shapeattribute already exists but targetshape differs

daz001:
When importing the sliderset and a shapeattribute already exists, it doesnt update the existing shape attribute, even though the targetshape (within a shapeattribute) may be different

At an initial glance, theres a few ways this could be fixed:
in SliderSet::Merge
   -always update the shape attribute (even if it already exists)
   -check if the shapetarget is different and update it
Or
   -clear the shape attribute from the slider set when the shape is deleted

This PR currently uses the last method as it seems to make sense to clear the old shape attribute when the old shape is deleted, though, I'm not sure if this would cover all scenarios - so it may be better to just already update a shape attribute if it exists. Also, I'm not sure if there is a reason that I didn't consider in which you might want to keep a shape attribute if the shape no longer exists

Test scenario (prior to PR)
1. Load CBBE Underwear
2. Delete all sliders
3. Delete CBBE
4. Load reference Template CBBE Body SMP (3BBB)

Note how the sliders don't affect the 'CBBE' shape. The issue is resolved with the PR

### DIFF
--- a/src/components/SliderSet.cpp
+++ b/src/components/SliderSet.cpp
@@ -273,8 +273,7 @@ void SliderSet::Merge(SliderSet& mergeSet, DiffDataSets& inDataStorage, DiffData
 
 	for (auto &s : mergeSet.shapeAttributes) {
 		// Copy new shapes to the set
-		auto it = shapeAttributes.find(s.first);
-		if (it == shapeAttributes.end()) {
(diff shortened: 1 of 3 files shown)


ousnius:
@daz001 There was very likely a reason for the shapeAttributes of the shape to not be removed, but I can't think of it right now and any tests I'm doing end up fine both before and after this change. So I will merge it for now.

PR #416, opened 2022-01-29
Cleaned up tool option user interface code

sts1skj:
The tool options are: x-mirror, connected-only, and all-selected-meshes (a.k.a. global brush collision).
The main simplification was to consolidate all of the UI-maintenance code in two functions.  In addition, I fixed a few things that were probably bugs:
- undiff, weight, color, and alpha had independent x-mirror options.
- all-selected-meshes would be toggled on when switching from the seg/part tab.
- x-mirror and connected-only were enabled for collapse-vertex, flip-edge, and split-edge
- connected-only was enabled on startup
- the text for all-selected-meshes was unclear


sts1skj:
Maybe the "Global Brush Collision" menu item should be renamed "Edit All Selected Meshes".  Maybe the variables and functions with "BrushCollision" in their names should be renamed.

Or maybe the variables and functions with "AllSelMesh" in their names that I added in this PR should be renamed to use "Global Brush Collision" instead.

sts1skj:
Instead of disabling the buttons in the vertical toolbar, it's possible to make them disappear completely.  Should I do so?

ousnius:
@sts1skj I thought about names and what to do with it for a while, but I'm currently of the opinion that the "Global Brush Collision" option can just be fully removed from the UI and brush logic (always on).

If you don't want to edit multiple meshes, then don't select multiple meshes. Any objections?

ousnius:
> Instead of disabling the buttons in the vertical toolbar, it's possible to make them disappear completely. Should I do so?

I'd prefer keeping them visible for now, it's less distracing seeing buttons pop up and away all the time.

sts1skj:
I agree that "Global Brush Collision" can be removed.  I'll delete it and commit a change.

ousnius:
@sts1skj Partition/segment editing intentionally turns off "Global Brush Collision" right now to prevent you accidentally selecting and editing multiple shapes at once.

Maybe with the feature removed, we need to add a warning for when you enter either of those tabs while having multiple shapes selected and prevent you from entering the tab while that's the case.

sts1skj:
Okay.  I'll add the warnings too.

sts1skj:
Would you like me to remove the "allMeshes" parameters of the functions GLSurface::CollideMeshes and GLSurface::UpdateCursor?  These parameters are now always set to "true".

ousnius:
@sts1skj It might make sense to keep them for the GL surface still. I don't think the brushes need it, though.

ousnius:
Replaced by #419

PR #419, opened 2022-02-02
Implemented Move Vertex tool

### LINKED ISSUE #302: Feature Request: Merge vertices together
Opposite to how you can split an edge and create a vertex, how about being able to merge 2 vertices together? This would be a huge help especially for outfits that have loose vertices and open holes because the author didn't merge them in Blender beforehand and having the option to merge vertices in Outfit Studio makes it so we don't have to go to Blender and re-export a nif. It'd be a major quality of life improvement imo and lets you fix various mistakes on a model if possible.

### LINKED ISSUE #380: Some vertice manipulation requests
I have 2 requests:

This not isn't very important, but it would be awesome, as you can see the move preserves the volume, it's perfect to move vertices without screwing up the shape, I believe the vertices slide across the edges.

ezgif com-gif-maker (8)

This is just a snap, you could do it like in the gif or a tool that just copy the translation and weights from one vertex to another, this would be good for fine tuning, fixing weights that cause seams and fixing mesh seams.

ezgif com-gif-maker (7)

### LINKED ISSUE #415: "Affect only one vertex" option for brushes
Currently, if you want to use an Outfit Studio brush (such as move, inflate, weight paint, or smooth) on exactly one vertex (or two if x-mirror is on), you have to adjust the Brush Size so that the brush is small enough to only hit one vertex.  You also have to click close enough to the vertex so that it's well inside the brush radius.  So I have to constantly adjust the Brush Size and precisely position the mouse pointer to get the various brushes to do what I want.

It would be nice if there were some option for the brushes that limited their area of effect to just one vertex (or two if x-mirror is on).

The usefulness of this option would vary by brush, depending on what proportion of the time I want to use that brush on just one vertex.  Here are my estimates per brush:
- move brush: 90%
- weight paint brush: 70%
- mask brush: 50%
- inflate/deflate brushes: 40%
- smooth and weight-smooth brushes: 40%
- undiff brush: 10%
- color and alpha brushes: I've never used these

Note that the mask brush already has a single-vertex alternative: the Select tool with the Vertex Edit option.

sts1skj:
The new tool has six tool options, five of which are new:
- X Mirror
- Weld
- Merge
- Restrict To Surface
- Restrict To Plane
- Restrict To Normal


sts1skj:
Some possibly unresolved stuff in this pull request:

I drew six new icons: one for the tool and the others for the tool options.  My icons are not necessarily good.

Should OutfitProject::IsVertexOnBoundary be in mesh instead and operate on mesh data instead of nif data?

I only added one key binding: Escape cancels an active vertex move.  Should there be more key bindings?

I made x-mirror mutually exclusive with weld and merge.  It would be messy to do a weld or merge with x-mirror on.  Should I implement it?

src/program/OutfitProject.cpp
```diff
@@ -3215,6 +3215,243 @@ bool OutfitProject::PrepareRefineMesh(NiShape* shape, UndoStateShape& uss, std::
...
+		newt.p3 = vCollapse[newt.p3];
+
+		uss.delTris.push_back(UndoStateTriangle{ti, t, part});
+		uss.addTris.push_back(UndoStateTriangle{ti, newt, part});
+	}
+}
+
+void OutfitProject::PrepareWeldVertex(NiShape* shape, UndoStateShape& uss, int selVert, int targVert, NiShape* targShape) {
+	// Get triangle data
+	uint16_t numVerts = shape->GetNumVertices();
```
ousnius: @sts1skj numVerts is unused here.

res/xrc/OutfitStudio.xrc
```diff
@@ -229,6 +235,46 @@
...
+								<tooltip>Merges the selected vertex into the target vertex, deleting it and attaching its triangles to the target vertex.</tooltip>
+								<bitmap>$(AppDir)\res\images\Merge.png</bitmap>
+								<toggle>1</toggle>
+								<checked>0</checked>
+								<disabled>1</disabled>
+							</object>
+							<object class="tool" name="btnWeld">
+								<label>Weld Vertex</label>
+								<tooltip>Welds the selected vertex to the target vertex, giving it the target's coordinates, coordinate slider diffs, and bone weights, but not UV.</tooltip>
+								<bitmap>$(AppDir)\res\images\Weld.png</bitmap>
```
ousnius, line 292: @sts1skj This file is missing from the commit.

ousnius (approved):

PR #421, opened 2022-02-04
Looping UI pool for SliderPanels

daz001:
The PR utilizes the idea of a looped ui pool specifically for slider panel ui. All none visible slider UI is now using a dummy slider panel (it has no additional ui components attached), the area that is visible in the scroll window is populated with real slider ui and the slider ui is repositioned and repopulated as there are events that would affect the scrolled window.

one thing i wasn't sure on how to handle is the SetFocus call to the sliderPanel, the current PR only calls it if the slider being focused is currently visible, otherwise i'm not sure how to handle it.

The intention of this is to reduce the number of user objects being used on windows in the hopes the window dragging wouldn't be delayed. This successfully reduces the number of user objects used by ALOT  (>50% when loading 100 sliders) and also reduces the startup time, due to the pool of visible sliders now being significantly lower by default. Unfortunately though, window dragging still appears delayed. 😒

Note. I haven't extensively tested everything


daz001:
Bugs:
- unloading project doesn't update visible sliders
- adding/deleting a slider doesn't updated visible sliders
- empty panels are added to fill the visible area if no sliders exist and the scrollwindow is updated
- additional sliders appear to be adding at the bottom of the sliderWindow

(Notes to self to fix before being considered for merging)

daz001:
I've fixed all the issues i can find and it may now be ready for consideration?

ousnius:
@daz001 Thanks for the effort and tests on this, however since it doesn't 100% fix the issue and adds a lot of code, I probably won't be merging this. I will think of a different or simpler workaround in the future.

PR #422, opened 2022-02-05
New cubic brush falloff formula; allow larger brush values to be entered (max 10.0)

### LINKED ISSUE #420: Move brush improvements
OS move brush feels weird, I don't know how the 3D softwares do it, but I took some screenshots as an exemple:

Zbrush | Softimage | OS

Untitled-1

Also, that's the maximum radius brush size in OS, it would be nice to have it extended.

sts1skj:


sts1skj:
Hopefully I caught every instance of "3.0f" in the brush size code and changed it.  Some of them were "3.000f" instead.  All of them were in OutfitStudio.h.  I didn't try to analyze that code to see if it made sense; I just changed the constant.

sts1skj:
I'm a little concerned that the minimum possible brush size, .001, may not be small enough with the new brush scale.  In "CBBE Body", there are several places where a brush size of .001 (with the new scale) gives a radius big enough that you can't isolate a point.  Maybe the brush-size code needs to be adjusted so that the resolution is .0001 instead of .001 .

ousnius (approved):

ousnius:
We decided to keep the brush size scale at 3 for now but increased the decimal places from 3 to 4. Additionally, you can manually enter values into the text fields up to 10.0 now.

PR #424, opened 2022-02-06
Adjusted scaling and defaults for brush strength for most brushes

sts1skj:
I changed the brush strength scaling so that now only inflate/deflate have a factor-of-10 scaling instead of all brushes.  I also increased the default value of brush strength by a factor of ten for smooth, move, undiff, weight, color, and alpha.  I also fixed some bugs that would make high strength values do nothing.

### DIFF
--- a/src/components/TweakBrush.cpp
+++ b/src/components/TweakBrush.cpp
@@ -171,8 +171,6 @@ TweakBrush::TweakBrush()
 	, inset(0.00f)
 	, strength(0.0015f)
 	, spacing(0.015f) {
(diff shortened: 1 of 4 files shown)


sts1skj:
I'm not sure I should have increased the default brush strength for undiff.

ousnius:
@sts1skj I think the changes are fine, but as you mention undiff should probably be 10x weaker again.

ousnius (dismissed):

ousnius (approved):

PR #434, opened 2022-02-18
Implemented Symmetrizer

### LINKED ISSUE #417: Mirror weights
A mirror weights tool, just a button actually, it be useful to fix weights that are not in perfect symmetry, it could also be useful to for the skinning process, so I wouldn't have to pose both sides to paint, I could just work on one side and then mirror the weights to the other side to speed up the process, given the topology is perfectly symmetrical of course. 

OS already can pick the mirrored bones, just need a button to mirror the weights from one side to another. Might need to specify from which to which like -X to +X and vice-versa.

sts1skj:
- Added command Mask Symmetric Vertices
- Added command Symmetrize Vertices
- Added command Mask Symmetric Triangles
- Fixed bug where applying/undoing/redoing a mesh edit would trash the masks.

### DIFF
--- a/res/xrc/Actions.xrc
+++ b/res/xrc/Actions.xrc
@@ -1282,4 +1282,334 @@
 			</object>
 		</object>
 	</object>
+	<object class="wxDialog" name="dlgVertexAsym">
+		<style>wxDEFAULT_DIALOG_STYLE|wxRESIZE_BORDER</style>
+		<size>423,155</size>
+		<title>Mask symmetric vertices</title>
+		<centered>1</centered>
+		<object class="wxBoxSizer">
+			<orient>wxVERTICAL</orient>
+			<object class="sizeritem">
+				<option>0</option>
+				<flag>wxALL|wxEXPAND</flag>
+				<border>5</border>
+				<object class="wxStaticText" name="maskSymmetricVerticesDescription">
+					<label>Masks all vertices that do not have any of the selected asymmetries.</label>
+					<wrap>600</wrap>
+				</object>
+			</object>
+			<object class="sizeritem">
+				<option>0</option>
+				<flag>wxEXPAND|wxALL</flag>
+				<border>5</border>
+				<object class="wxFlexGridSizer">
+					<rows>0</rows>
+					<cols>3</cols>
+					<vgap>0</vgap>
+					<hgap>0</hgap>
+					<growablecols>1</growablecols>
+					<growablerows></growablerows>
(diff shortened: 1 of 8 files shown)


sts1skj:
I didn't try to implement a triangle symmetrizer.  Maybe I should have.

sts1skj:
This symmetrizer does something so natural and intuitive that it can be described with a single English word: symmetrizer.  But the implementation is horribly messy.  This pull request is 1500 lines of code!  It's so complex that my programming instincts are strongly suggesting that there's something I could have done to make this code simpler.  But I'm not seeing it right now.  Sigh.

sts1skj:
I've decided I need to do a better job separating the symmetrization options for a bone pair.

sts1skj:
I have an idea for a way to simplify this code a little bit: extract the mask handling from MatchSymmetricVertices and FindVertexAsymmetries.  Maybe I'll try it tomorrow.

sts1skj:
Though the last commit had a net increase of 7 code lines, it also breaks functionality up into more functions, which is longer but simpler, so this last commit is overall a significant simplification.

Should the symmetrizer code be split out of OutfitProject.{h,cpp} into a separate unit, say Symmetrizer.{h,cpp}?  Should it be pulled out of the OutfitProject class and put into its own class?  (I'm leaning towards "no" for both of these questions.)

ousnius:
Thanks for the work on this and the additional cleanup + code comments.

> Should the symmetrizer code be split out of OutfitProject.{h,cpp} into a separate unit, say Symmetrizer.{h,cpp}? Should it be pulled out of the OutfitProject class and put into its own class? (I'm leaning towards "no" for both of these questions.)

Right now, I'm also leaning towards no. The OutfitStudio and OutfitProject sources are very large and they definitely need to be split up into multiple sources at some point, but that's a bigger project and not important now.

I will be trying out the symmetrizer myself. Provided I don't find any issues, would you think this is ready to merge?

sts1skj:
Yes, I think it's ready to merge, assuming you find no issues and don't see anything that could be cleaned up or improved.

ousnius:
@sts1skj I've given it a quick try, the feature is great and I couldn't find any crash or bug. It only took me a short while to understand that the dialog shows checkboxes for asymmetric features only (and once they're symmetric they're gone).

This is probably the most useful for sliders and bone weights. Since it's a new separate feature/dialog and likely won't break anything existing, I will merge it.

ousnius (approved):

PR #437, opened 2022-02-24
Implemented new smoothing brush algorithms

### LINKED ISSUE #253: Feature request: improve smoothing brushes
**What is it that you want to do that isn't currently possible?**
Smoothing brushes don't do what I want in several different circumstances:

1.  When the mesh connections are asymmetrical, the result is unbalanced in favor of the side with more mesh connections.  (By mesh connection, I mean a triangle edge connecting it to another vertex.)

2.  The position-smoothing brush often moves the vertex position more along the surface than perpendicular to it.  Usually, but not always, such along-the-surface motion is undesirable.

3.  The position-moving brush often moves vertices in or out of the surface in addition to moving them along it.  Sometimes this is desirable; sometimes it isn't.

**Request the feature**
1.  Make all vertices in a connected neighborhood of the vertex matter to smoothing.  Unfortunately, this raises a parameter issue: how big should the neighborhood be?  Should the size of the neighborhood be specified by vertex count or by geometric distance?  Should the user be able to specify the neighborhood size?

If it were undesirable to add another brush parameter slider, a crude solution could be to just use the eight closest connected vertices (or ten, or twelve, or whatever).  Another crude solution would be to use the existing brush-size parameter to also control neighborhood size (which might work well or might work horribly).

It might also be nice (if the user had the option to specify a large neighborhood with lots of vertices) if the smoothing function used weighted averaging, with farther vertices affecting the result less than closer vertices.

2.  It would be nice if the position-smoothing brush had an option to restrict the movement to be perpendicular to the surface.  This could be a user-controlled switch.  A switch to restrict motion the other way, to be tangent to the surface, might (or might not) also be useful.

One question is how to determine the direction perpendicular to the surface.  A simple way would be to just average the normal directions of the adjacent triangles.  But I think you'd get better results if you did a simple plane fit (possibly weighted) of all the vertices in the neighborhood.

3.  One possibility is to add a three-way switch controlled by the user that affects how the move-position brush works: option 1 is free movement; option 2 is along-the-surface movement; and option 3 is perpendicular-to-the-surface movement.

But another possibility would be to determine which option to use automatically: if the angle between the surface normal and the camera direction is less than some angle (say 30 degrees), the motion would automatically be restricted to be along the surface; if the angle were greater than some angle (say 60 degrees), the motion would automatically be restricted to be perpendicular to the surface.

sts1skj:
For the position-smoothing brush, the new algorithm is a "balanced pair circle fit" smoother.  It finds balanced pairs of adjacent points and fits a circle through each pair.

For the mask-smoothing and weight-smoothing brushes, the new algorithms are "balanced pair parabola fit" smoothers.  They find balanced pairs of adjacent points and fit a parabola through each pair.

I think the circle fitter works well.  The parabola fitter doesn't work great, it seems to me, but is still probably better than the existing hclap code.

### DIFF
--- a/res/xrc/Actions.xrc
+++ b/res/xrc/Actions.xrc
@@ -1282,4 +1282,335 @@
 			</object>
 		</object>
 	</object>
+	<object class="wxDialog" name="dlgVertexAsym">
+		<style>wxDEFAULT_DIALOG_STYLE|wxRESIZE_BORDER</style>
+		<size>423,155</size>
+		<title>Mask symmetric vertices</title>
+		<centered>1</centered>
+		<object class="wxBoxSizer">
+			<orient>wxVERTICAL</orient>
+			<object class="sizeritem">
+				<option>0</option>
+				<flag>wxALL|wxEXPAND</flag>
+				<border>5</border>
+				<object class="wxStaticText" name="maskSymmetricVerticesDescription">
+					<label>Masks all vertices that do not have any of the selected asymmetries.</label>
+					<wrap>600</wrap>
+				</object>
+			</object>
+			<object class="sizeritem">
+				<option>0</option>
+				<flag>wxEXPAND|wxALL</flag>
+				<border>5</border>
+				<object class="wxFlexGridSizer">
+					<rows>0</rows>
+					<cols>3</cols>
+					<vgap>0</vgap>
+					<hgap>0</hgap>
+					<growablecols>1</growablecols>
+					<growablerows></growablerows>
+					<object class="spacer">
+						<option>0</option>
+						<flag>wxLEFT|wxRIGHT</flag>
+						<border>5</border>
+						<size>0,0</size>
+					</object>
+					<object class="sizeritem">
+						<option>0</option>
+						<flag>wxLEFT|wxRIGHT</flag>
+						<border>5</border>
+						<object class="wxStaticText" name="currentUnmaskedLabel">
+							<label>Vertices currently unmasked:</label>
+							<wrap>-1</wrap>
+						</object>
+					</object>
+					<object class="sizeritem">
+						<option>0</option>
+						<flag>wxLEFT|wxRIGHT|wxALIGN_RIGHT</flag>
+						<border>5</border>
+						<object class="wxStaticText" name="currentUnmaskedText">
+							<label>0</label>
+							<wrap>-1</wrap>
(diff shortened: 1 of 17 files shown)


sts1skj:
Some questions:
1. Should the user be able to pick which smoothing algorithm they want?  Currently they can't.  (I suggest no.)
2. Should using balanced pairs be a tool option that applies to any of the three smoothers?  Any of the existing lap or hclap smoothers could use balanced pairs as an option.  (I suggest no.)
3. The new smoothers use a lap-style update instead of an hclap-style update.  (hclap-style updating tugs the results back towards the starting values, sort of.)  Should I rewrite them to use hclap-style updating?  (I suggest no.)
4. I put the functions ParabolaFit, CircleFitMidpoint, and CircleFitNearestPoint in TweakBrush.cpp.  These functions could reasonably be moved elsewhere.  ParabolaFit relies on Mesh functions, so it could be easily transferred into Mesh.{h,cpp}, but moving it to nifly's Object3d.{h,cpp} would be more messy, though definitely doable.  CircleFitMidpoint and CircleFitNearestPoint could easily be moved to Mesh.{h,cpp} or Object3d.{h,cpp}.  Are these functions okay in TweakBrush.cpp, or should they be moved?
5. The new mask and weight smoothers are not great.  Should they be dropped in favor of the existing hclap smoothers?

sts1skj:
One thing I noticed while experimenting with the new bpcf position-smoother is that it sometimes produces slightly poor results near points with connections like this:
image
The reason is that the bpcf code relies on normals calculated by Mesh::SmoothNormals, and that function doesn't produce great results when a point's triangles are unbalanced on one side, like the point in the image above.  It produces a normal for the central point that is more dominated by the upper triangles' normals than the lower triangles' normals, because there are four upper triangles and only two lower triangles.  So the normals at points like this aren't quite right.  I'm not sure, but I think I've noticed this a few times while playing Skyrim.

So maybe Mesh::SmoothNormals could be improved.  One idea is to do a plane fit through balanced pairs of neighboring vertices.  Another idea is calculating triangle normals for a set of temporarily generated triangles that only use balanced pairs of neighboring vertices (instead of using the mesh's own neighboring triangles).  Should I look into this?

ousnius:
@sts1skj There's been some changes in dev (adjacent points) that require a rebase/merge.

PR #441, opened 2022-03-04
Various bone features and fixes such as checking for bad bones

### LINKED ISSUE #379: Copy special "cloth" boneweight works incorrectly
I tested this problem with different versions and found that 4.9.1 is latest properly working version for this issue
next version available for download (5.0.7) have this bug

So when you copy boneweight from some clothing that have special cloth bones:
they are copied and have physics in game, but they are **dispalced** in game (see img 1) and **not displaced** in nifskope (see img 2)
But for proper custom boneweight copy they **should be dispalced** in nifskope (see img 3) and in result will look good in game
https://imgur.com/a/bHpdX6a

Or maybe I understand something wrong, what I do:
1)load project with my new dress (my dress is img 2)
2)import from nif another working dress (custom working dress is img 3) with custom bones
3)set reference to working dress and copy boneweight for my new dress
4)delete custom dress, save and build outfit

### LINKED ISSUE #406: "Apply Pose To Mesh" doesn't update slider diffs
If you select a pose with at least one bone rotated and then click the "Apply To Mesh" button in the Pose frame, the model's vertex's positions are updated, but the slider diffs for those vertices are not.  So the slider diffs now offset the vertices in the wrong direction.

For example:
1.  Load "CBBE Body".
2. Turn on posing, select the bone "NPC R Thigh", and pick an X rotation of 1.67 (about 90 degrees), so the body kicks its right leg out horizontally.
3. Click "Apply To Mesh".
4. Set the "Chubby Legs" slider to 200%.

The vertical left leg is appropriately thickened, but the vertices of the horizontal right leg are moved along the leg, so it doesn't get significantly thicker or thinner.
image

### LINKED ISSUE #440: Custom bones don't appear right with "Show Bones" or "Show Nodes"
Custom bones don't appear at all when you turn on "Show Bones".  They do appear when you turn on "Show Nodes", but their location doesn't update when you change them with "Edit Bone".

sts1skj:
- Added command, "Check for bad bones", that opens a dialog that lists bones with inconsistent transforms and gives the user options for fixing the problems.
- Added a "Bad Bones" submenu to the bone context menu with two options for fixing bad bones.
- Display broken-bone symbol next to bad bones in Bones tab.
- Fixed custom bones not appearing with "Show Bones" on.
- Fixed incorrect bone display in FO3/FONV with "Show Bones" on.
- Made "Show Bones" show posed bones when "Show Pose" is on.
- Added scale to the pose parameters.
- Removed old bone-scale slider and code.
- Made "Apply Pose To Mesh" undo-able.
- Made "Apply Pose To Mesh" update position diff sliders too.
- Added display of vertex coordinates in status bar with the weight brush.


src/program/OutfitStudio.cpp
```diff
@@ -1078,11 +1078,17 @@ OutfitStudioFrame::OutfitStudioFrame(const wxPoint& pos, const wxSize& size) {
 	if (outfitBones) {
 		wxBitmap noneImg(wxString::FromUTF8(Config["AppDir"]) + "/res/images/icoNone.png", wxBITMAP_TYPE_PNG);
 		wxBitmap changeImg(wxString::FromUTF8(Config["AppDir"]) + "/res/images/icoChange.png", wxBITMAP_TYPE_PNG);
+		wxBitmap brokenBoneImg(wxString::FromUTF8(Config["AppDir"]) + "/res/images/icoBrokenBone.png", wxBITMAP_TYPE_PNG);
+		wxBitmap brokenBoneChangeImg(wxString::FromUTF8(Config["AppDir"]) + "/res/images/icoBrokenBoneChange.png", wxBITMAP_TYPE_PNG);
 		wxImageList* boneStateImages = new wxImageList(16, 16, false, 2);
```
ousnius: Initial size of the image list can be increased from 2 to 4.

ousnius (changes requested):

ousnius (dismissed):

ousnius (approved):

PR #456, opened 2022-04-27
Fixes for wxWidgets-gtk3 3.1.6

sts1skj:
- With gtk3 (and not gtk2), newer versions of wxWidgets tend to make the
function glewInit return the error GLEW_ERROR_NO_GLX_DISPLAY unless you
initialize things in exactly the right sequence, which can be messy.  GLEW
still seems to work if you ignore the error, though, so I added code to ignore
the error.
- Added ToStdString in a dozen or so places (where implicit conversion from
wxString to std::string was used) so that you don't have to build wxWidgets
with the --enable-stl option.
- Updated the linux build instructions, mostly to improve clarity.

### DIFF
--- a/Readme-linux.txt
+++ b/Readme-linux.txt
@@ -2,15 +2,16 @@ Building BodySlide and Outfit Studio on Linux
 
 There's just one file that specifies the build process: CMakeLists.txt.
 
-Install wxWidgets 3.1.3 or newer.  Use the "--enable-stl" option to
-configure.  If you get errors about an ABI mismatch, that means you
-compiled wxWidgets with a different compiler version than BS&OS (and
-wxWidgets is strangely picky about that).  Either gtk2 or gtk3 works.
-With gtk2, you have more background color problems; with gtk3, many
-widgets are distorted because they don't have enough space.  Note that
-many of wxWidget's configure options (such as --enable-universal)
-result in a broken wxWidgets library, so prefer to use as few options
-as possible.
+Install wxWidgets 3.1.3 or newer.  If you get errors about an ABI
+mismatch, that means you compiled wxWidgets with a different compiler
+version than BS&OS.  Either gtk2 or gtk3 works.  With gtk2, you have
+more background color problems; with gtk3, many widgets are distorted
+because they don't have enough space.  Note that many of wxWidget's
+configure options (such as --enable-universal) result in a broken
+wxWidgets library, so prefer to use as few options as possible.
+Also note that some distribution-provided wxWidgets packages are
+broken, so it's likely that you'll have to build wxWidgets yourself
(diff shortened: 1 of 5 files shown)


ousnius:
@sts1skj ``wxString::ToStdString()`` is a wide (UTF-16) string on Windows iirc, which is not what I want (codebase expects UTF-8). Can you use ``wxString::ToUTF8()`` instead?

sts1skj:
According to the wxWidgets documentation, `wxString::ToStdString()` returns a `std::string`, not a `std::wstring`.  The function `wxString::ToStdWstring()` returns a `std::wstring`.

I picked `wxString::ToStdString()` in an effort to be consistent with the rest of the code.  I saw many places in OutfitStudio.cpp where `wxString::ToStdString()` was used to convert from `wxString` to `std::string`.  But there are also many places where `ToUTF8().data()` is used.  According to the wxWidgets documentation, `ToStdString` uses locale encoding, while `ToUTF8` uses UTF8 encoding.  I don't know what the possibilities are for locale encoding, so I don't understand the difference between these two conversion methods.

One thing is clear from the wxWidgets documentation: when you convert `wxString` to `std::string` implicitly, it's the same as calling `ToStdString`.  Since all of the code I changed was doing implicit conversion before, my replacing the implicit conversions with calls to `ToStdString` should not change the result.

Do you want me to use `ToUTF8().data()` instead of `ToStdString()` for the 16 changes in this PR?  What about the many existing places where `ToStdString()` was called before this PR?

ousnius:
> One thing is clear from the wxWidgets documentation: when you convert wxString to std::string implicitly, it's the same as calling ToStdString. Since all of the code I changed was doing implicit conversion before, my replacing the implicit conversions with calls to ToStdString should not change the result.

In that case, the use of ``ToStdString()`` is fine for now for this PR. I wasn't fully aware of that.

In the future (separate to this PR), all occurences of ``ToStdString()`` where encoding can matter should be replaced by ``ToUTF8()`` (or ``utf8_str()``) when converting from wxString to std::string and ``FromUTF8()`` when converting from std::string to wxString (from logic to UI) respectively.

Most of the code already does this, I did a big parse back then for Chinese file names (UTF-16 on Windows) and UTF-8 file contents (XML files) back then. It's possible there are important gaps here, though, hence it should be thoroughly looked at once more.

ousnius (approved):

sts1skj:
This PR only fixes the implicit conversions from `wxString` to `std::string` that were generating errors with g++-10 when wxWidgets is built without the "--enable-stl" option.  Outfit Studio has many more implicit conversions in both directions between `wxString` and `std::string`.  I don't know how we could find them all.  There could be thousands.

PR #545, opened 2025-06-14
Optimize shape zapping when generating morphs

miere43:
`shapeZapIndices` contains ranges of vertices to delete, instead of deleting vertices one by one calculate continuous range and delete vertices in big chunks.

In my tests this change has improved batch build time by 20% (single-thread performance).

<details><summary>Before (00:03:01)</summary>
<p>

image

</p>
</details> 

<details><summary>After (00:02:26)</summary>
<p>

image

</p>
</details> 

Profiling results:

<details><summary>Before (39.69 s)</summary>
<p>

image

</p>
</details> 

<details><summary>After (1.93 s)</summary>
<p>

image

</p>
</details> 

ShapeZapIndices zone in traces above refers to this part of code:

```c++
for (int i = shapeZapIndices.size() - 1; i >= 0; i--)
    verts.erase(verts.begin() + shapeZapIndices[i]);
}
```

### DIFF
--- a/src/program/BodySlideApp.cpp
+++ b/src/program/BodySlideApp.cpp
@@ -1091,6 +1091,8 @@ bool BodySlideApp::WriteMorphTRI(const std::string& triPath, SliderSet& sliderSe
 		if (shapeZapIndices.size() > 0 && shapeZapIndices.back() >= shapeVertCount)
 			continue;
 
+		auto zapRanges = FindContinuousRanges(shapeZapIndices);
+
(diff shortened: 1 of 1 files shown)


ousnius (approved):

ousnius:
@miere43 For me the batch build went from 0.14s to 0.12s in x64 release build.

I've changed the code so the zap range would only be determined once and I've also made it apply to UV sliders.
Feel free to check if that improved performance even further for you.

I've done a binary comparison of the output before and after the changes and the files are identical.

Thanks for the contribution.

PR #547, opened 2025-06-18
Remove Mesh::BuildTriAdjacency

miere43:
Data generated by this method was not used anywhere (stored in `vertTris` variable). Speeds up batch build time a little bit.

### DIFF
--- a/src/components/Mesh.cpp
+++ b/src/components/Mesh.cpp
@@ -31,25 +31,6 @@ std::shared_ptr<AABBTree> Mesh::CreateBVH() {
 	return bvh;
 }
 
-void Mesh::BuildTriAdjacency() {
-	if (!tris)
-		return;
-
-	vertTris = std::make_unique<std::vector<int>[]>(nVerts);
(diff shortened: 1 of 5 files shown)


ousnius:
I'm confused by ``vertTris`` not being used, but I guess it's true.

However this can't possibly speed up batch build time, since the Mesh class isn't used when building outfits in any way. It's only used for rendering meshes in the preview window and Outfit Studio.

miere43:
> However this can't possibly speed up batch build time, since the Mesh class isn't used when building outfits in any way. It's only used for rendering meshes in the preview window and Outfit Studio.

Yes, you are correct. I mean that preview window now opens a little bit faster 🙂

ousnius:
`vertTris` used to be used in several functions in Mesh.cpp, but not anymore.
https://github.com/ousnius/BodySlide-and-Outfit-Studio/blob/46c408e60d2fbe66e886bb6db34d81f327bb288c/src/components/Mesh.cpp

ousnius (approved):

PR #548, opened 2025-06-22
fix compile issues on arch linux

wrb2012:


lib/FSEngine/FSBSA.cpp
```diff
@@ -31,7 +31,11 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ***** END LICENCE BLOCK *****/
 
 #include "FSBSA.h"
+#ifdef __linux__
+#include <directx/dxgiformat.h>
+#elif
```
ousnius: The #elif with a missing condition is causing a compile error on MSVC. Can you change it to an else?

wrb2012:
Thanks!

src/render/GLSurface.cpp
```diff
@@ -708,7 +708,7 @@ void GLSurface::SetSize(uint32_t w, uint32_t h) {
 	if (!SetContext())
 		return;
 
-	glViewport(0, 0, w, h);
+	glViewport(0, 0, w*this->canvas->GetContentScaleFactor(), h*this->canvas->GetContentScaleFactor());
```
ousnius, line 711: The docs of wxWidgets are saying:
> For example, in a [wxGLCanvas](https://docs.wxwidgets.org/latest/classwx_g_l_canvas.html) created with the size of 100 (logical) pixels, the rightmost physical pixel coordinate will be 100*GetContentScaleFactor() under all platforms.

So this should be correct.

However, I'm on a 4K monitor with 150% scaling (Windows 11) myself and didn't need to apply the scale factor on my system and I'm wondering why.

It looks like that's because my system does DPI scaling and GTK does content scaling:
```cpp
double wxWindowBase::GetContentScaleFactor() const
{
    // By default, we assume that there is no mapping between logical and
    // physical pixels and so the content scale factor is just 1. Only the
    // platforms that do perform such mapping (currently ports for Apple
    // platforms and GTK 3) override this function to return something
    // different.
    return 1.0;
}
```

In any case, looks like this can be merged without any side effects.

ousnius:
Thanks for the contribution.
Is there anything else you want to do with the PR? Otherwise I will merge it.

ousnius (approved):

PR #554, opened 2025-08-02
improve icons with HIDPI

wrb2012:
I don't have raw icons, and I've tried some AI tools to upscale PNGs, but not satisfied. So this is just a small try to improve HIDPI support.


ousnius:
Thank you for the effort. Since the list of icons is incomplete and there's no source anymore, I won't merge for now.

PR #566, opened 2026-03-10
Performance: meshoptimizer, slider optimizations, texture prefetch

SulfurNitride:
## Summary
- **meshoptimizer vertex cache optimization**: Integrates [meshoptimizer](https://github.com/zeux/meshoptimizer) library to reorder render indices for GPU vertex cache efficiency. Applied automatically in `Mesh::CreateBuffers()` and after partition changes in Outfit Studio.
- **ApplySliders fast-path**: Skip sliders with empty `linkedDataSets` or zero value, avoiding unnecessary iteration in the inner build loop.
- **ApplyReferenceNormals per-shape**: Refactored to operate on a single shape instead of iterating all shapes, avoiding redundant file lookups during builds.
- **Texture prefetch**: New `ResourceLoader` prefetch infrastructure reads texture file data in a background thread (file I/O only, no GL context needed). `LoadTexture()` checks the prefetch cache first for fast GL uploads. Called on outfit selection and app startup in BodySlide, and on project/outfit load in Outfit Studio.

## Test plan
- [ ] Build on Windows (MSVC) — verify meshoptimizer compiles and links for both BodySlide and OutfitStudio
- [ ] Open Outfit Studio, load a project — meshes should render correctly with optimized index order
- [ ] Open BodySlide, select an outfit, open preview — textures should load (prefetched in background)
- [ ] Build single outfit and batch build — verify output .nif files are identical to before
- [ ] Verify no regressions in slider application or reference normals

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


ousnius:
Thanks, but I don't want to add a library/dependency for the mesh optimization. Did you measure how much of an impact optimizing the mesh for rendering has?

As for the texture prefetch – it's a good idea but I don't agree with the current implementation. One reason is because it requires loading the project NIF an additional time and introduces a lot of additional code. Same as above, what's the impact?

If you want to work on the PR and remove the mesh optimizer and do some other tweaks, that would be cool. If not, that's fine too.

SulfurNitride:
The texture prefix, runs in the background upon the outfit loading. So for the user there is no noticeable increase to latency or waiting on the program. As for mesh optimizer it was just something I was trying out, and it only reduced the outfits build time by about 5 seconds for 5000 outfits. So there isn't really to much of an improvement there and I can see how you would see it as pointless. 

But the main reason for the prefetch was that for some reason upon loading it up at least on my linux build, the preview window would take about 3-5 seconds to load and then all other outfits would load within a second. With the fix, I noticed on my end I noticed it was all under a second. So its up to you on that.

ousnius:
> The texture prefix, runs in the background upon the outfit loading. So for the user there is no noticeable increase to latency or waiting on the program. As for mesh optimizer it was just something I was trying out, and it only reduced the outfits build time by about 5 seconds for 5000 outfits. So there isn't really to much of an improvement there and I can see how you would see it as pointless.
> 
> But the main reason for the prefetch was that for some reason upon loading it up at least on my linux build, the preview window would take about 3-5 seconds to load and then all other outfits would load within a second. With the fix, I noticed on my end I noticed it was all under a second. So its up to you on that.

I think I would prefer only including the `ApplySliders` and `ApplyReferenceNormals` in the baseline for now to keep code complexity from increasing.

Feel free to make a new PR or edit the existing one whenever you can spare the time. If not let me know, that's fine too. Thanks. :)

PR #567, opened 2026-03-10
Fix category clearing on profile switch, reload data after settings change

SulfurNitride:
## Summary
- **Fix stale categories**: `LoadCategories()` was clearing the category collection internally, which wiped base categories when loading from additional paths. Moved the clear to `LoadAllCategories()` so it happens once before any loading, and added a `Clear()` method to `SliderCategoryCollection`.
- **Clear stale SelectedOutfit**: When switching target game or settings, the previously selected outfit may no longer exist. `LoadData()` now detects this and clears the selection rather than leaving a stale reference.
- **Reload data after settings change**: After saving settings in the Settings dialog, the app now calls `LoadAllCategories()`, `LoadAllGroups()`, `LoadSliderSets()`, and `LoadData()` so changes take effect immediately without restarting. Also updates `targetGame` so subdirectory scanning picks up the correct paths.

## Test plan
- [x] Open Settings, change target game, click OK — outfit list and categories should update immediately
- [x] Switch between different game targets — no stale categories should appear from previous game
- [x] Select an outfit, switch to a game where that outfit doesn't exist — selection should clear gracefully
- [x] Verify no regressions in normal startup and outfit selection flow

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/components/SliderCategories.cpp
+++ b/src/components/SliderCategories.cpp
@@ -7,8 +7,6 @@ See the included LICENSE file
 #include "../utils/PlatformUtil.h"
 
 int SliderCategoryCollection::LoadCategories(const std::string& basePath) {
-	categories.clear();
-
 	wxArrayString files;
 	wxDir::GetAllFiles(basePath, &files, "*.xml");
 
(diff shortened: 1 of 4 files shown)


ousnius (approved):

ousnius:
Thanks for the contribution.

PR #582, opened 2026-05-31
BodySlide: Add a 'Build Logbook' feature

Togekiss:
logbook

# Build Logbook (name open to change, of course)

## Why
I have a lot of outfit and preset mods that I tinker with constantly. Keeping track of what I already built, with which preset, and what I didn't built yet was a challenge- manually writing them down or adding them to groups is slow and error-prone. So I looked at the code.

- `Log_BS.txt` is not meant to be persistent, and while it could be parsed by an external script, it doesn't have all the information I wanted. 
- `BuildSelection.xml` is persistent, but it only stores decisions about zaps and preferences when there's mesh conflicts- it doesn't reflect what has been actually _built_.

So I created a new log file. At first I just wanted it to be a simple list with "outfit name, preset name, has edits yes/no", so I could play around with BS, and parse that log file when I'm done to create groups for each preset. But I got carried away, and ended up with something that might be useful for more people!

## What
A new `BuildLogbook.xml` file that should be a record of BodySlide outputs to the game's mod folders (or the designated Output Path):
- Whenever an outfit is built, be it on its own or as a batch build, an entry is created in the Logbook.
- If an outfit with the same mesh path as a previous one is built, its Logbook entry is _replaced_ with the newest info.
- If an outfit is deleted from BodySlide using the ALT key, its Logbook entry is also deleted.

*Note: This does not scan the game folders for previous BodySlide builds, or is aware of anything that changed outside BodySlide.*

Additionally, a text label in the bottom right of the BodySlide UI shows if the selected outfit has been built.
- Hovering over it shows information about the last build
- Right-clicking allows users to load the latest build back to BodySlide, or to remove the entry from the Logbook (but the mesh stays in the game/output folder)

### BuildLogbook.xml file
```xml
<BuildLogbook>
    <!-- A simple entry that lists a Mesh (identified by its path), which set it was built with, and which preset was used. -->
    <Mesh path="meshes\armor\studded\female\boots" set="DH 3BA Leather boots" preset="CBBE Curvy"/>

    <!--
    An entry that describes a Mesh, and stores information about any sliders that have been modified.
    Only Sliders that don't match the default values of the set and the preset are listed here.
    -->
    <Mesh path="meshes\armor\imperial\f\cuirasslight" set="DH 3BA Imperial light armor" preset="CBBE Curvy (Outfit)">
        <!-- Example modifications of the set's Zaps -->
        <SetSlider name="Stockings" size="big" value="100"/>
        <SetSlider name="Stockings" size="small" value="100"/>
        <!-- Example modifications of the preset's sliders-->
        <SetSlider name="BreastsPressed_v2" size="big" value="49"/>
    </Mesh>
</BuildLogbook>
```

## How

My C++ is a bit rusty and it's the first time I work with wxWidgets, but I tried to copy the way other features are implemented as much as possible.

### BodySlide.xrc
The box containing the `BuildSelection` choice was reduced from stretching through 3 columns to occupying 2.
In that freed up space, I added a `wxStaticText` Label that shows if the current outfit has been built or not.
I also added the right-click menu near the end of the file.

### BuildLogbook.h/cpp
This module contains the classes BuildLogbook and BuildLogbookFile. Heavily inspired from `BuildSelection.h/cpp`, they mostly create, populate and hold the data structure and XML file respectively, and convert it back and forth. They don't contain heavy logic.

### BodySlideApp.h/cpp
Where all the magic happens!

As this is a hefty module, I have tried to keep new functions close to their BuildSelection siblings to ease navigation.

When the app loads, `InitBuildLogbook()` reads `BuildLogbook.xml` and loads it to a `BuildLogbook` object that will persist during execution. \*
*Note: Check 'Known Issues' section below.*

Then, and whenever a new outfit or a new preset are selected, `UpdateLogbookLabel()` checks if there is a matching Logbook entry, and sets up the text, tooltip and right-click menu accordingly. The text color follows the same logic as BuildSelection: grey if there's no entry, yellow if there is an entry but it's not the current one, and blue if the entry matches the currently loaded set and preset.

When the "Build" button is clicked, `SaveSetToLogbook()` or `RemoveSetFromLogbook()` are called, depending on the ALT key.
`SaveSetToLogbook()` runs near the end of `BuildBodies()` if everything was built correctly. It creates a new `BuildLogbookEntry` and stores the mesh path, set name and preset we're building with. With `CalculateLogbookSliders()`, it checks the `sliderManager` sliders to find out if the preset was modified before building. This allows to save micro-adjustments (i.e. fixing nipples poking through cloth, outfits with squeeze sliders) without having to save a preset variant only for that outfit. This also saves the actual Zap values the outfit was built with (vs the last selected ones in BuildSelection)
After saving or removing an entry from the BuildLogbook, the XML file is updated.

*Note: I know there are similar slider calculations in the code, but I wanted to disturb existing functions as little as possible, so I chose to make a small one that can be plugged in with one line.*

To respect "Batch Build" parallelism, `CalculateLogbookSliders()` only runs once at the start (since the whole batch should be using the same Preset and slider values)\*\*. Then, `BuildLogbookEntries` (or OutputPath strings, if we're batch deleting) are created and stored in a separate vector inside `batchBuildMutex`. The main BuildLogbook and the XML file are only updated at the end of the batch build operation.
*Note: Check 'Known Issues' section below.*

If "Load the latest build" is selected in the label's right-click menu, `ActivateLogbookEntry()` is called. This selects and loads the last recorded Outfit and Preset (and also updates the outfit/preset/favorite buttons selection UI at the top of the screen). If the entry contains any slider data, it applies the changes to the affected sliders. The Preview and the Logbook label are refreshed when it finished loading everything.

## Known issues
As a draft, this feature is not perfect, complete, or 100% robust. Here is a list of things I am currently aware of. Some could be fixed, some will be fixed (even if just for my personal use), and some might be left as a skill issue.

- \*Speedy, not robust against external changes to the XML during executions
  -  I'm aware `BuildSelection` opens, loads and closes the file constantly. But I thought doing that only to check `hasEntry()` would be a waste, especially with a file that can grow very big. I'm careful with the file, but this should be re-evaluated if this were to be made public
- \*\*In Batch Builds, if the currently loaded outfit has a Zap, it gets applied to all entries regardless if they even have that zap
  - Well, if an entry has a slider that isn't found in the outfit/preset, it simply doesn't get loaded. But it's still an annoying bug I have to fix, plus it _can_ mess with entries if they so happen to have the same Zap name. I have to test how much calculating sliders for every outfit impacts performance.
- ~~Inconsistent behavior when loading Zaps that conflict with BuildSelection~~
  - Fixed in 8c8c268cbc66f47e220fb02d5d394715ecad51c3.
- As mentioned, it's not aware of any changes made outside BodySlide, or when this feature isn't active.
  - Very likely out of scope. We'd have to instruct the user to delete or back up their BuildLogbook.xml at their convenience.
- Outfit or preset renames don't update the Logbook, so those won't be detected as entries anymore.
  - Acceptable 'skill issue'. Easy enough for a mid-level user to open the .xml and "Find and replace" the affected items
- If a preset is re-saved with different slider values, the Logbook won't know that. Loading an entry will load the preset with its current values, not with the values it was actually built with.
  - Not sure if this is a good or a bad thing. Let's assume people will create a different preset if they want a new body shape.
  
## Possible improvements
The current state of this feature (minus the bugs) is enough for my personal use, but some things that could be added are:

- A toggle in the Settings / Config file / CLI argument to enable or disable this feature
- A checkbox next to the Logbook Label to automatically load the Logbook entry when a Set/outfit that has one is loaded
  - It would load the Preset and apply slider edits. Loading/reloading a preset shouldn't trigger it, though.
- When Batch building, an option to apply Logbook entries to any outfit with a matching one
  - BuildSelection would still take priority to decide which Set to build for each Mesh. Then we'd attempt to find a Logbook entry for that Set.
  - This would be useful to quickly re-apply a Logbook with lots of micro-adjustments. Could be useful when reinstalling mods, or if a modder wants to share their zap and squeeze refinements for a certain SliderPreset and Outfit pack combination.
- Adding translations and Automation support
  - I'm not familiar with those, so I'd only look into them if this feature is useful enough to be merged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@
 *.vs
 
 /build
+/.vscode
--- a/BodySlide.vcxproj
+++ b/BodySlide.vcxproj
@@ -826,6 +826,7 @@
     <ClCompile Include="src\components\Anim.cpp" />
     <ClCompile Include="src\components\Automorph.cpp" />
     <ClCompile Include="src\components\BuildSelection.cpp" />
+    <ClCompile Include="src\components\BuildLogbook.cpp" />
     <ClCompile Include="src\components\ClippingFixer.cpp" />
     <ClCompile Include="src\components\DiffData.cpp" />
     <ClCompile Include="src\components\Mesh.cpp" />
--- a/BodySlide.vcxproj.filters
+++ b/BodySlide.vcxproj.filters
@@ -1899,6 +1899,9 @@
     <ClCompile Include="src\components\BuildSelection.cpp">
       <Filter>Components</Filter>
     </ClCompile>
+    <ClCompile Include="src\components\BuildLogbook.cpp">
+      <Filter>Components</Filter>
+    </ClCompile>
     <ClCompile Include="src\components\ClippingFixer.cpp">
       <Filter>Components</Filter>
     </ClCompile>
--- a/BuildLogbook.xml
+++ b/BuildLogbook.xml
@@ -11,13 +11,13 @@ If a mesh gets deleted, its entry in this file will be removed.
 <BuildLogbook>
 
     <!--
-    A simple entry that lists a Mesh (identified by its path), which set it was built with, and which preset was used. 
+    A simple entry that lists a Mesh (identified by its path), which set it was built with, and which preset was used.
     Each "path" should be unique across the file.
     This simple entry did not register any change to the default values of the set or preset sliders.
     -->
     <Mesh path="meshes\armor\studded\female\boots" set="DH 3BA Leather boots" preset="The Female Gaze v2 Outfit"/>
 
-    <!-- 
+    <!--
     An entry that describes a Mesh, and stores information about any sliders that have been modified.
     Only Sliders that don't match the default values of the set and the preset should be listed.
     -->
--- a/BuildLogbook.xml
+++ b/BuildLogbook.xml
@@ -1,28 +1,33 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<!-- A file to keep track of the currently built meshes -->
+<!--
+This file should represent the contents of the game/ouput folder by keeping track of each mesh that has been built with BodySlide.
+
+Every time a mesh is built, either on its own or as part of a batch build, an entry will be created in this file.
+If a mesh that already exists in this file is rebuilt, its entry will be updated.
+If a mesh gets deleted, its entry in this file will be removed.
+
+-->
 <BuildLogbook>
 
-    <!-- 
-    A simple entry that lists the mesh, which set it was built with, and which preset was used.
-
-    There should be only one "Mesh" entry per path.
-    If a mesh is built again, the old entry should be deleted and replaced with the new one, 
-    so that the logbook represents the contents of the game/ouput folder.
+    <!--
+    A simple entry that lists a Mesh (identified by its path), which set it was built with, and which preset was used. 
+    Each "path" should be unique across the file.
+    This simple entry did not register any change to the default values of the set or preset sliders.
     -->
     <Mesh path="meshes\armor\studded\female\boots" set="DH 3BA Leather boots" preset="The Female Gaze v2 Outfit"/>
 
     <!-- 
-    A "Mesh" can contain slider information.
-    Ideally, this should be limited to sliders that were modified from the set's and preset's default values.
+    An entry that describes a Mesh, and stores information about any sliders that have been modified.
+    Only Sliders that don't match the default values of the set and the preset should be listed.
     -->
     <Mesh path="meshes\armor\imperial\f\cuirasslight" set="DH 3BA Imperial light armor" preset="The Female Gaze v2 Pushup">
-        <!-- For example, modifications of the set's Zaps -->
+        <!-- Example modifications of the set's Zaps -->
         <SetSlider name="Outfit var.1" size="big" value="100"/>
         <SetSlider name="Outfit var.1" size="small" value="100"/>
         <SetSlider name="Outfit var.2" size="big" value="0"/>
         <SetSlider name="Outfit var.2" size="small" value="0"/>
-        <!-- Or edits of the preset's sliders-->
+        <!-- Example modifications of the preset's sliders-->
         <SetSlider name="BreastsPressed_v2" size="big" value="24"/>
     </Mesh>
 </BuildLogbook>
--- a/BuildLogbook.xml
+++ b/BuildLogbook.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- A file to keep track of the currently built meshes -->
+<BuildLogbook>
+
+    <!-- 
+    A simple entry that lists the mesh, which set it was built with, and which preset was used.
+
+    There should be only one "Mesh" entry per path.
+    If a mesh is built again, the old entry should be deleted and replaced with the new one, 
+    so that the logbook represents the contents of the game/ouput folder.
+    -->
+    <Mesh path="meshes\armor\studded\female\boots" set="DH 3BA Leather boots" preset="The Female Gaze v2 Outfit"/>
+
+    <!-- 
+    A "Mesh" can contain slider information.
+    Ideally, this should be limited to sliders that were modified from the set's and preset's default values.
+    -->
+    <Mesh path="meshes\armor\imperial\f\cuirasslight" set="DH 3BA Imperial light armor" preset="The Female Gaze v2 Pushup">
+        <!-- For example, modifications of the set's Zaps -->
+        <SetSlider name="Outfit var.1" size="big" value="100"/>
+        <SetSlider name="Outfit var.1" size="small" value="100"/>
+        <SetSlider name="Outfit var.2" size="big" value="0"/>
+        <SetSlider name="Outfit var.2" size="small" value="0"/>
+        <!-- Or edits of the preset's sliders-->
+        <SetSlider name="BreastsPressed_v2" size="big" value="24"/>
+    </Mesh>
+</BuildLogbook>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,6 +317,7 @@ set(BSOS_OUTFIT_STUDIO_SOURCES
 set(BSOS_BODYSLIDE_SOURCES
 	${BSOS_COMMON_SOURCES}
 	src/components/BuildSelection.cpp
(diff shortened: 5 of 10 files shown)


Togekiss:
### Conflict between BuildSelection and BuildLogbook

Before:
* User loads an outfit with zaps, all set to false by default
* User selects zap 1 and 2 and builds an outfit
  * BuildSelection saves zap choices 1 and 2 as true
  * BuildLogbook saves modifications for zap 1 and 2
* User deselects zap 1 and selects zap 3, but doesn't build
  * BuildSelection saves zap choice 1 as false, 2 and 3 as true
* User loads the latest Logbook entry
  * The preset gets reactivated and BuildSelections get reapplied, setting 1 as false and 2 and 3 as true
  * BuildLogbook gets applied, setting 1 and 2 as true. But it doesn't attempt to edit zap 3, so it stays as true

After commit 8c8c268cbc66f47e220fb02d5d394715ecad51c3 :
* User loads the latest Logbook entry
  * The preset gets reactivated _without loading any BuildSelection zap choices_
  * BuildLogbook is applied against the default set/preset values, and only zap 1 and 2 are set to true
* User loads any outfit or preset, including reloading the currently selected ones
  * Any changes that loading the BuildLogbook entry introduced are wiped
  * BuildSelection is loaded, and sets the previously selected values of 1=false, 2=true, 3=true
  
**Loading a Logbook entry that contains zap edits does not modify or update BuildSelection**
This is on purpose, to make loading a Logbook entry as inoffensive as possible- loading it only "previews" it back to BS, and discarding it is as easy as re-loading the preset. If a user wants to make that permanent, they can re-select the zaps by choice.
But this is open to change!

ousnius:
@Togekiss I like the idea of a build log. Thanks for the detailed explanation and draft.

TL;DR: Keep the functionality but use the existing BuildSelection.xml file and expand it instead.

Right now there are two concerns the logbook handles:
1. Storing what set was built to where using what preset
2. Storing slider values that differ from the preset

In my opinion, the first concern should be handled by changing BuildSelection to include the preset in the `<OutputChoice />` element and by always adding an `<OutputChoice />` element for built sets.

That also has the benefit of your last build becoming the conflict winner automatically when you add more mods to your install later on that write to the same output path but didn't exist before. The new label in the UI is great and can use the BuildSelection then.

By using BuildSelection.xml for the first concern, that would leave the new BuildLogbook.xml storing only the slider values deviating from the preset (correct me if I'm wrong). Adding a new file "just" for that that seems weird. The BuildSelection.xml already has `<ZapChoice />` elements. Maybe we can just add `<SliderValue />` elements there? This also solves the zap conflicts between two files.

The code would have to be merged and BuildSelection loaded/handled in more places.

Togekiss:
@ousnius thanks for the review!

Merging everything into BuildSelection makes a lot of sense! The concept of "last selection" and "last build" are very similar, and I bet they will be the same in 90% of the cases, but they're juuuuust different enough to keep it in mind.

This implementation was meant to disturb existing functionalities as little as possible. But if you like the idea, I can work on a version that's more integrated with everything else!

_But_ it would conflict with the current behavior of BuildSelection. Right now:
- User selects "Outfit A" as the default for that mesh
  - "Outfit A" is stored as the `<OutputChoice />` in `BuildSelection.xml`
- User loads and builds "Outfit B", without changing the checkbox to make it the new default
  - "Outfit B" is recorded in `BuildLogbook.xml`, but the `<OutputChoice />` in `BuildSelection.xml` is still "Outfit A"
  
Or, as the zap conflict issue reflected, the last zaps that a user selected in the UI might not be the same zaps the outfit was built with.

----

**Some suggestions to merge it with existing files:** Let me know what you think, prefer, or add any more suggestions!

For `<OutputChoice />` in `BuildSelection.xml`:
- If the entry doesn't have a `preset` attribute, it means it's only a selection, but it was not built yet.
- If the entry has a `preset`, it meants it was built at some point.
- If an outfit is deleted (With Build+ALT), or with "Forget the latest build", only its `preset` attribute is removed, and the  `<OutputChoice />` stays as a default build option.
- When an outfit is built, it's stored as an `<OutputChoice />` with a `preset`, regardless of if it has conflicting meshes or not.
  - *Should it overwrite any previous `<OutputChoice />` with that mesh and become the new default?*
  - *Whan happens to that entry if a user selects another Set as the default one?*
    - Overwriting it would defeat the purpose of having a build log that says "This is what you have in your game right now:"
    - The only options I can think of are to add some other flag to `<OutputChoice />`s, or to simply not allow to change the default one without rebuilding. I don't like either of them, any suggestions?

For `<ZapChoice />` in `BuildSelection.xml`:
- If the selected outfit doesn't have a recorded build, `<ZapChoice />`s work as usual- stored every time the user changes them.
- If the selected outfit was built, changing Zaps in the UI doesn't update `BuildSelection.xml`.
  - They only update when a user builds again.
  - This means that outfits will be pre-loaded with the zaps the outfit was last built with, not with the last UI choice. I don't think many people would mind that difference...
  
  
About `<SliderValue />`

As mentioned, the goal is to keep track of micro-adjustments (like fixing stubborn nipple clipping), and outfit-specific  sliders- without having to save a whole new preset for each outfit.

We could add those to `BuildSelection.xml` (which IMO, makes sense for outfit "squeeze" sliders and the like- they're in the Zap section of the UI after all)

And/or, we could leverage the `set` attribute of `<SliderPresets>`, which right now doesn't do much.
- *Saving* a Preset saves it to a `/SliderPresets` XML file as usual, with no / an empty `set` attribute.
- When *building* an outfit, any modified sliders get saved under `<Preset name="XX" set="YY">`
  - I'm not sure if it'd be easier/better to store the whole thing, or only the modified sliders. The whole thing is more convenient to save and load, but it's less file-size and human-readable friendly...
- When loading a preset, the default slider values (the entry without a `set` attribute) are loaded.
- When selecting "Load the latest build", the preset entry with the matching `set` is loaded.
  - (Or the set-specific modifications could be loaded automatically- but I like to keep the option of loading the default preset or the tweaked variant at will)
- When a build choice is deleted or forgotten, the `set`-specific SliderPresets could be deleted- or not. Honestly, if they're not loaded automatically and there's no `BuildSelection` to activate the "Load this" button, keeping them there wouldn't be a problem.

If it's implemented carefully, this shouldn't change the way Presets work at all.

_But_ it would 'pollute' the SliderPreset files that modders share. If we went this route, I'd store these in a `MySliderPresetName_variants.xml` file or something... maybe someone would like to ship their preset with hand-tuned squeeze sliders for every TAWOBA piece, haha

PR #588, opened 2026-06-07
BodySlide: Use game's data path if there's no project files in appdir

### LINKED ISSUE #585: BodySlide on Linux doesn't see any bodies/outfits in Skyrim Special Edition
**Description**
When I run BodySlide native Linux version it doesn't see any bodies/outfits. I've tried to run it both from Fluorine Manager and without it (just copied CBBE mod directly in the Data directory for testing)

Image

**How to reproduce**
1. Build BodySlide on Arch Linux
2. Install some mods like CBBE
3. Add Skyrim Special Edition as game
4. Run BodySlide

RevengeRip:
If there's no `SliderSets` folder next to BodySlide executable it will use `Data/CalienteTools/BodySlide` directory as project path. Otherwise, it will fall back to appdir.

Fixes #585

### DIFF
--- a/BodySlide.vcxproj
+++ b/BodySlide.vcxproj
@@ -793,6 +793,7 @@
     <ClInclude Include="src\utils\ConfigurationManager.h" />
     <ClInclude Include="src\utils\Log.h" />
     <ClInclude Include="src\utils\PlatformUtil.h" />
+    <ClInclude Include="src\utils\ProjectUtil.h" />
     <ClInclude Include="src\utils\StackTrace.h" />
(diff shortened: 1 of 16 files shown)


ousnius:
For other games like FO4 the folder can be "Tools" instead of "CalienteTools".

RevengeRip:
Done. ~It now also checks for `SliderSets` directory inside either `CalienteTools` or `Tools` in case they're just empty directories with no project files~.

EDIT: that wasn't necessary since we're already looking for `BodySlide` directory.

ousnius:
Thanks for the contribution. I added some changes and refactors.

PR #590, opened 2026-06-11
BodySlide: Fix zaps overwritten during Batch Build if they have the same name

Togekiss:
Fixes bug described in https://github.com/ousnius/BodySlide-and-Outfit-Studio/issues/589

Skip applying the values loaded in the UI (sliderManager) to Zaps when batch building.
All the zap modifications, even for the outfit loaded in BodySlide when starting the batch build, are stored in `BuildSelection.xml`. So there is no need to load zaps from `sliderManager` at all.

This avoids accidentally applying wrong zap values to all outfits that have a zap with the same name.

(This PR also fixes a small conversion mismatch that `clangd` complained about)


ousnius:
Thanks for the contribution. :)

PR #595, opened 2026-06-30
Add option to Hide Conflict Losers in Outfit/Body list

Dramacydal:
Adds a new checkable filter option to the outfit search context menu that hides outfits which lost a resolved output file conflict, showing only winners and unresolved outfits.

Didn't touch translations since they seem to be generally outdated

image


ousnius:
Looks good to me, I have no remarks. Thanks for the contribution.

PR #596, opened 2026-06-30
Add option to hide virtual shapes in preview

Dramacydal:
Add option to hide virtual shapes (used for HDT/SMP collisions) from preview

image
image

### DIFF
--- a/src/components/Mesh.h
+++ b/src/components/Mesh.h
@@ -127,6 +127,7 @@ public:
 	std::shared_ptr<AABBTree> bvh = nullptr;
 
 	bool bVisible = true;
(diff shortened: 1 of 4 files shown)


ousnius:
Instead of using the name "virtual" for detection, you can make the checkbox hide shapes that have no shader or have the hidden flag set.

This is also how the clipping fixer detects these collision shapes that are not rendered in-game (you can check out the code there).

ousnius:
Maybe in BodySlide/PreviewPanel, these shapes without shaders should just be hidden always/by default? And the checkbox could be for showing them instead.

Dramacydal:
@ousnius did as you suggested

ousnius:
Works perfectly. Thanks for the contribution.

PR #600, opened 2026-08-05
Fixed Linux wrong launch path for OutfitStudio

PetricaT:
On the new Linux native build, if the user tries to launch Outfit Studio from within BodySlide, the application will silently fail and this error will appear in the terminal 

execvp(**/home/USER/Desktop/BodySlide-and-Outfit-Studio**OutfitStudio) failed with error 2!

Where: 
 - **/home/USER/Desktop/BodySlide-and-Outfit-Studio** - is where i  saved the compiled results locally
 - **OutfitStudio** - is the binary name.
 
 The fix is simple, just add a slash before the executable call
 
 Tested and worked locally.

### DIFF
--- a/src/program/BodySlideApp.cpp
+++ b/src/program/BodySlideApp.cpp
@@ -1326,7 +1326,7 @@ void BodySlideApp::LaunchOutfitStudio(const wxString& args) {
 #ifdef _WIN32
 	const wxString osExec = "OutfitStudio.exe";
 #else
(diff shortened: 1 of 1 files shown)


ousnius:
Thanks for the contribution.

I changed it to use wxFileName, which is even better. Let me know if that works for you also.

PetricaT:
Tested the latest artifact build, works perfectly now.

